PR #6, opened 2015-01-21
Replace shatar with shazar

hseeberger:


.gitignore
```diff
@@ -3,3 +3,6 @@ typesafe_conductr_cli.egg-info
 __pycache__
 deb_dist/*
 !deb_dist/Dockerfile
+.idea
+*.iml
```
2m, line 7: I tend to add IDE specific ignores to the global ignore file. Saves lots of typing. :)

typesafe_conductr_cli/shazar.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
```
2m, line 1: I know this was in the previous file, but we do not need this anymore. Executable file is being generated by pip.

typesafe_conductr_cli/shazar.py
```diff
@@ -0,0 +1,64 @@
...
+                    name = os.path.join(source_base_name, os.path.relpath(path, start=args.source))
+                    zip_file.write(path, name)
+        else:
+            zip_file.write(args.source, source_base_name)
+
+    dest = shutil.move(
+        temp_file,
+        os.path.join(args.output_dir, "{}-{}.zip".format(source_base_name, create_digest(temp_file)))
+    )
+    print('Created digesgted ZIP archive at {}'.format(dest))
```
2m: s/digesgted/digested/

2m:
LGTM after a couple of minor nitpicks.

.gitignore
hseeberger, line 7: Yeah, and therefore these patterns weren't present for me which led to adding them to Git accidentally.

typesafe_conductr_cli/shazar.py
hseeberger, line 1: If it doesn't hurt, let's keep it, because this way we can easily execute the script for testing purposes.

typesafe_conductr_cli/shazar.py
hseeberger: Good catch!

Issue #59, opened 2015-09-11
Lots of blank lines for conduct log


huntc:
Messages appear with lots of blank lines in between - what's happening?

![screen shot 2015-09-11 at 5 08 23 pm](https://cloud.githubusercontent.com/assets/694893/9808998/c318a5f6-58a7-11e5-83c8-e4c6ae1bc1f6.png)

Here's the corresponding json:

``` json
[
    {
        "host": "dc9f9cab9411",
        "message": "Set conductr index pattern as default\n",
        "timestamp": "2015-09-11T07:00:35.814Z"
    },
    {
        "host": "dc9f9cab9411",
        "message": "{\"_index\":\".kibana\",\"_type\":\"config\",\"_id\":\"4.1.2\",\"_version\":1,\"created\":true}",
        "timestamp": "2015-09-11T07:00:35.85Z"
    },
    {
        "host": "dc9f9cab9411",
        "message": "Refresh conductr index pattern\n",
        "timestamp": "2015-09-11T07:00:35.851Z"
    },
    {
        "host": "dc9f9cab9411",
        "message": "{\"_shards\":{\"total\":1,\"successful\":1,\"failed\":0}}",
        "timestamp": "2015-09-11T07:00:35.882Z"
    },
    {
        "host": "dc9f9cab9411",
        "message": "module.js:340\n    throw err;\n          ^\n",
        "timestamp": "2015-09-11T07:00:35.94Z"
    },
    {
        "host": "dc9f9cab9411",
        "message": "\n",
        "timestamp": "2015-09-11T07:00:35.94Z"
    },
    {
        "host": "dc9f9cab9411",
        "message": "Error: Cannot find module 'lodash'\n    at Function.Module._resolveFilename (module.js:338:15)\n    at Function.Module._load (module.js:280:25)\n    at Module.require (module.js:364:17)\n    at require (module.js:380:17)\n    at Object.<anonymous> (/tmp/53dad3668b1cb31432ae853bb61de656/53dad3668b1cb31432ae853bb61de6565adba4e270167ca5c13066abcfdf2f04/conductr-kibana-0.1.0/src/config/index.js:1:71)\n    at Module._compile (module.js:456:26)\n    at Object.Module._extensions..js (module.js:474:10)\n    at Module.load (module.js:356:32)\n    at Function.Module._load (module.js:312:12)\n    at Module.require (module.js:364:17)\n",
        "timestamp": "2015-09-11T07:00:35.943Z"
    },
    {
[...]

PR #65, opened 2015-10-05
Remove blank lines

markusjura:
This PR addresses the issue #59.

Blank lines are now stripped with `rstrip()` for the commands:
- logs
- info
- services
- events


edwardcallahan:
LGTM

huntc:
The CLI should not wrap text when being used with "more" or "grep" for example. Is this behaviour preserved?

markusjura:
The log messages are now getting trimmed (similar to sbt-conductr).

PR #74, opened 2015-11-10
Color code terminal Warning and Error output

markusjura:
Improvement of the terminal output text `Error:` and `Warning:`.

Before an error message was looking like this:

```
ERROR: My error message
```

Now, the error message looks like this, whereas the `Error:` word is red and underlined (can't be displayed here in Markdown):

```
Error: My error message
```

ANSI Codes are used which are support on Unix (incl. OS X) systems and Windows: http://stackoverflow.com/questions/287871/print-in-terminal-with-colors-using-python

### DIFF
--- a/conductr_cli/ansi_colors.py
+++ b/conductr_cli/ansi_colors.py
@@ -1,0 +1,4 @@
+RED = '\033[91m'
+YELLOW = '\033[93m'
+UNDERLINE = '\033[4m'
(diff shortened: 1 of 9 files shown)


markusjura:
This follows the color code convention of `brew`.

markusjura:
screen shot 2015-11-10 at 11 37 34

huntc:
Nice.

PR #75, opened 2015-11-11
Implement Control Protocol v2

fsat:
Fixes https://github.com/typesafehub/conductr/issues/925.
- [x] Rename v1.1 Control Protocol to v2
- [x] Present `bundle.conf` found within bundle zip file as part of the `conduct load` request as opposed to parsing the `bundle.conf` for fields declared within the file.
- [x] Present the overlay `bundle.conf` if found within the configuration zip in the similar fashion


conductr_cli/test/test_conduct_load_v2.py
```diff
@@ -0,0 +1,114 @@
...
+            patch('sys.stderr', stderr), \
+            patch('conductr_cli.bundle_utils.zip_entry', zip_entry_mock):
+            args = self.default_args.copy()
+            conduct_load.load(MagicMock(**args))
+
+        self.assertEqual(
+            as_error(strip_margin("""|Error: Problem with the bundle: Unable to find bundle.conf within the bundle file
+                                     |""")),
+            self.output(stderr))
+        zip_entry_mock.assert_called_with('bundle.conf', self.bundle_file)
```
huntc, line 197: None of the base test methods being called here. Is that expected (why have the base here then)?

fsat, line 197: We still need the base since there are a number of success test cases which are dependent on the base class.

conductr_cli/conduct_load.py
```diff
@@ -95,15 +95,42 @@ def get_v_1_0_payload(bundle_name, bundle_file, bundle_configuration):
...
+        configuration_file, bundle_conf_overlay = (None, None)
+        if args.configuration is not None:
+            print('Retrieving configuration...')
+            configuration_name, configuration_url = get_url(args.configuration)
+            configuration_file, configuration_headers = urlretrieve(configuration_url)
+            bundle_conf_overlay = bundle_utils.zip_entry('bundle.conf', configuration_file)
+
+        files = [('bundleConf', ('bundle.conf', bundle_conf))]
+        if bundle_conf_overlay is not None:
+            files.append(('bundleConfOverlay', ('bundle.conf', bundle_conf_overlay)))
```
huntc, line 116: Don't we want to add the conf overlay part still, even if there's no overlay conf (in which case it is just a part with no filename)?

conductr_cli/test/test_conduct_load_v2.py
```diff
@@ -0,0 +1,197 @@
...
+            conduct_load.load(MagicMock(**args))
+
+        self.assertEqual(
+            open_mock.call_args_list,
+            [call(self.bundle_file, 'rb'), call(config_file, 'rb')]
+        )
+
+        expected_files = [
+            ('bundleConf', ('bundle.conf', 'mock bundle.conf')),
+            ('bundle', ('bundle.zip', 1)),
```
huntc, line 155: So we took the path of supporting the absence of the second part then?

fsat, line 155: Yes we did, it's implemented this way within ConductR.

conductr_cli/conduct_load.py
fsat, line 116: No - ConductR is able to handle the absence of the conf overlay part.

huntc:
LGTM

fsat:
Squashed commits - manual tested successfully against ConductR master branch.

fsat:
Also tested successfully against ConductR 1.0 branch.

Issue #72, opened 2015-11-02
Shorthand for bundle expressions


huntc:
Support a short-hand expression for referring to bintray releases in a similar vein to Docker. For example:

   conduct load https://bintray.com/artifact/download/typesafe/test-bundle-repo/typesafe/simple-test/v0-2aa7b82c2a8db90712252048857f61f9042a404eee058bfc72832daae2342bb7/simple-test-v0-2aa7b82c2a8db90712252048857f61f9042a404eee058bfc72832daae2342bb7.zip

...should be able to abbreviated to something generic:

```
conduct load typesafe/test-bundle-repo/simple-test:v0-2aa7b82c2a8db90712252048857f61f9042a404eee058bfc72832daae2342bb7
```

...and further:

```
conduct load test-bundle-repo/simple-test
```

The above uses the following convention to address the bundle:

```
address = [ "urn:x-bundle:" ] , [ organization , "/" ] , [ repository , "/" ] , package , [ ":" , "v" , compat-version [ "-" , digest ] ]
```

The CLI shall receive a resolver style of plugin architecture supporting the following behaviours.

If no scheme is provided then `urn:x-bundle:` will be assumed (we should register this with IANA).

If the `organization` is not supplied then a resolver will attempt to resolve it against the organisations it has been configured for (we support the name "typesafe" by default in our supplied resolvers).

If the `repository` is not supplied then a resolver will attempt to resolve it against the repositories it has been configured for (we support the name "bundle" by default in our supplied resolvers).

If the `compat-version` (compatibility version) is not supplied then latest upload is assumed and the resolvers will resolve the latest bundle within the repositories they are associated with, addressing the archive with the latest digest that was uploaded.

So the above load command can be shortened further:

```
conduct load typesafe/test-bundle-repo/simple-test
```

...which would get the latest.  Also:

```
conduct load simple-test
```

...would resolve to typesafe/bundle/simple-test given the default resolvers supplying "typesafe" and "bundle" for organisation and repository names respectively.

The following:

```
conduct load typesafe/test-bundle-repo/simple-test:v0
```

would get the latest uploaded bundle with a compatibility version of "v0".
### Bintray considerations

[...]

PR #78, opened 2015-11-18
Implement Resolver mechanism within ConductR CLI

fsat:
Part of the task required to implement #72 
- [x] Encapsulate existing file and http url resolution into uri-based resolver
- [x] Implement caching mechanism
- [x] Implement bintray-based resolver supporting bundle short-hand expression (as per #72)

### DIFF
--- a/conductr_cli/conduct.py
+++ b/conductr_cli/conduct.py
@@ -8,6 +8,9 @@ import os
 
 DEFAULT_PORT = os.getenv('CONDUCTR_PORT', '9005')
 DEFAULT_API_VERSION = os.getenv('CONDUCTR_API_VERSION', '1')
+DEFAULT_CLI_SETTINGS_DIR = os.getenv('CONDUCTR_CLI_SETTINGS_DIR', '{}/.conductr'.format(os.path.expanduser('~')))
+DEFAULT_BUNDLE_RESOLVE_CACHE_DIR = os.getenv('CONDUCTR_BUNDLE_RESOLVE_CACHE_DIR',
+                                             '{}/cache'.format(DEFAULT_CLI_SETTINGS_DIR))
 
 
 def add_ip_and_port(sub_parser):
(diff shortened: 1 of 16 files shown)


fsat:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;i7nnHd2j&#x2F;138-shorthand-for-bundle-expressions"> Shorthand for bundle expressions</a>

conductr_cli/conduct.py
```diff
@@ -8,6 +8,9 @@
 
 DEFAULT_PORT = os.getenv('CONDUCTR_PORT', '9005')
 DEFAULT_API_VERSION = os.getenv('CONDUCTR_API_VERSION', '1')
+DEFAULT_CLI_SETTINGS_DIR = os.getenv('CONDUCTR_CLI_SETTINGS_DIR', '{}/.conductr'.format(os.getenv('HOME')))
```
huntc: Will that work on Windoze?

conductr_cli/resolvers/bintray_resolver.py
```diff
@@ -0,0 +1,153 @@
...
+            print('Bintray credentials loaded from {}'.format(BINTRAY_CREDENTIAL_FILE_PATH))
+            return username, password
+
+
+def bintray_download_url(bintray_username, bintray_password, org, repo, package_name, compatibility_version, digest):
+    if compatibility_version is None and digest is None:
+        # Get latest version
+        package_endpoint = '{}/packages/{}/{}/{}'.format(BINTRAY_API_BASE_URL, org, repo, package_name)
+        package = get_json(package_endpoint, bintray_username, bintray_password)
+        latest_version = package['latest_version']
```
huntc, line 72: Unfortunately I've seen this field set to null having published just one version. I don't think we should rely on it. Instead, we should always get the package and iterate through the `latest-v` property names to determine which one is the latest.

conductr_cli/resolvers/bintray_resolver.py
```diff
@@ -0,0 +1,153 @@
...
+            raise BintrayResolutionError(
+                'Unable to find version for owner={} repo={} package={} version={}'.format(
+                    org, repo, package_name, bintray_version))
+        else:
+            path = matching_files[0]['path']
+            download_url = '{}/{}/{}/{}'.format(BINTRAY_DOWNLOAD_BASE_URL, org, repo, path)
+            return download_url
+
+
+def get_version_timestamp(bintray_username, bintray_password, org, repo, package_name, version):
```
huntc: I don't think that this function is used anymore.

conductr_cli/resolvers/bintray_resolver.py
```diff
@@ -0,0 +1,153 @@
...
+        response = requests.get(uri, auth=(username, password))
+    else:
+        response = requests.get(uri)
+    response.raise_for_status()
+    return json.loads(response.text)
+
+
+def log(message, org, repo, package_name, compatibility_version, digest):
+    if compatibility_version is None and digest is None:
+        print('{} org={} repo={} bundle={} version={}'.format(message, org, repo, package_name, 'latest'))
```
huntc: We should use the shorthand convention here i.e.

{org}/{repo}/{package_name}:latest

...same for the others below.

conductr_cli/resolvers/bintray_resolver.py
fsat, line 72: @huntc - I've implemented the change to infer latest version from list of attribute names that starts with `latest-v` below

PR #81, opened 2015-11-27
Add -q switch to conduct load command

fsat:
This change would mean `conduct load` will only display bundle id upon the activation of `-q` switch.

Replace all the `println()` with python's logging library, ensuring same behaviour and output.

Utilize python logging's message formatter to provide warning and error message formatting (i.e. colours and underlines).


huntc:
LGTM

huntc:
Is this still a WIP? Looks ready to merge.

PR #83, opened 2015-11-30
Ensure sandbox command works against Python 3.2

fsat:
Detect if `FileNotFoundError` is present (it is for python 3.3 & above), else trap `OSError` instead (which is raised in place of `FileNotFoundError` in python 3.2 and below)


huntc:
LGTM

Issue #85, opened 2015-12-02
Downloading and caching resilience


huntc:
If a download fails via the resolver (CTRL-C halfway through the download), then the CLI will attempt to load the half-downloaded file from the cache - and fail.

We should not write directly to the cache, but instead write to a temporary file and then atomically move it.

fsat:
Thanks for catching this - I'll look into this

fsat:
Closed given #86 is merged.

PR #86, opened 2015-12-02
Prevent partial downloaded bundle to be resolved

fsat:
This is done by ensuring the download is stored into a temp file which is moved to the actual path once downloaded is successful


fsat:
Fixes #85

PR #93, opened 2015-12-17
Resolving default ip address from docker machine should handle CalledProcessError

fsat:


fsat:
Still needs manual test with docker machine

huntc:
LGTM

PR #110, opened 2016-02-22
Ignore non key/value pair lines

### LINKED ISSUE #107: bintray credential parsing
Apparently bintray credential file formats can screw up our parsing of them. Our parser will have to become more sophisticated. From @henrikengstrom:

> I haven’t changed the credentials file lately and it contained more than `user` and `password`, e.g. `realm` and `host`.

huntc:
For example, comments.

Fixes #107

### DIFF
--- a/conductr_cli/resolvers/bintray_resolver.py
+++ b/conductr_cli/resolvers/bintray_resolver.py
@@ -5,12 +5,14 @@ from requests.exceptions import HTTPError
 import json
 import logging
 import os
+import re
 import requests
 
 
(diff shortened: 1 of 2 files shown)


markusjura:
The current code would fail if the `value` would contain a `=` sign, e.g.

```
secret = ah5k=sdh
```

We should split the line by key value only by the first `=` sign.

huntc:
Good catch! I shall fix tomorrow!

fsat:
Lgtm - nice catch @markusjura !

Issue #115, opened 2016-03-08
getLevelName issue with logging


huntc:
From a customer report:

---

By the way, a minor point I encountered when upgrading:

I discovered that Python 3.4.1 (what I had been using) does not work with the latest ConductR CLI due to (from https://docs.python.org/3.4/library/logging.html#module-logging https://docs.python.org/3.4/library/logging.html#module-logging):

logging.getLevelName(lvl)
...
Changed in version 3.4: In Python versions earlier than 3.4, this function could also be passed a text level, and would return the corresponding numeric value of the level. This undocumented behaviour was considered a mistake, and was removed in Python 3.4, but reinstated in 3.4.2 due to retain backward compatibility.

I switched to Python 3.4.2 and that fixed the problem.

fsat:
Fixed by #116

PR #116, opened 2016-03-09
Obtain logging level from python's logging module constant

fsat:
This is done instead of calling `getLevelName` method which doesn't work in Python 3.4.1


fsat:
Fixes #115

PR #130, opened 2016-03-24
Remove the store_true action from wait-timeout

JustinPihony:
Also fixed the help message.


huntc:
Is the default to store anyhow? By removing the arg you're just cleaning up the code?

Can you please describe the motivation for the PR as the title isn't doing it for me. Thanks.

JustinPihony:
No, the default is to store, not store_true. See [the docs](https://docs.python.org/2/library/argparse.html) and find store_true for the documentation. This makes it so that what should be expecting an int is expecting no arguments and defaulting to true. 

So, when you run `conduct load --wait-timeout 60` it looks at the 60 as a different argument

huntc:
Thanks for this.

PR #137, opened 2016-06-21
Support the new Docker for Mac and Windows

### LINKED ISSUE #135: Support the new Docker for Mac
Is not working correctly with the beta version of Docker for Mac. They've changed things again on us it seems. From a customer report, it appears that there is no "default" machine anymore. Using --ip works though.

huntc:
The new Docker for Mac and Windows environments use a local hypervisor instead of docker-machine/boot2docker. As such we now check whether we can run docker at all prior to docker-machine/boot2docker. If we can then we need not check any further, and can assume docker's IP to be the loopback address.

Fixes #135

### DIFF
--- a/conductr_cli/host.py
+++ b/conductr_cli/host.py
@@ -8,15 +8,19 @@ from subprocess import CalledProcessError
 def resolve_default_ip():
     def resolve():
         try:
-            return with_docker_machine()
-        except validation.NOT_FOUND_ERROR:
+            terminal.docker_ps()
+            return '127.0.0.1'
+        except CalledProcessError:
             try:
-                return with_boot2docker()
+                return with_docker_machine()
             except validation.NOT_FOUND_ERROR:
                 try:
-                    return with_hostname()
+                    return with_boot2docker()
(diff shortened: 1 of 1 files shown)


huntc:
Merged sans review as things are broken given the new Docker release.

markusjura:
LGTM

manuelbernhardt:
Reporting from Mac OS Sierra with the docker toolbox, which is to say docker-machine.

It appears that somehow the IP of the virtual machine gets resolved to loopback whereas the machine runs on its own IP (returned correctly via `docker-machine ip`). I don't know that this is in any way related to Sierra though. I think the problem may be in the logic [here](https://github.com/typesafehub/conductr-cli/blob/d22a0bba4d987d8151dbf4172ba5a644c973ff5d/conductr_cli/host.py#L11) where it is assumed that if `docker ps` runs from the terminal, then loopback is a valid IP. In reality though this may not be the case. I would suggest checking first for docker-machine, then for boot2docker and finally default to loopback.

PR #139, opened 2016-06-29
WIP: DON'T MERGE: Supporting docker 1.12

markusjura:


conductr_cli/docker.py
```diff
@@ -0,0 +1,28 @@
...
+    else:
+        version = terminal.docker_version()
+        if version >= 1.12 and not os.getenv('DOCKER_MACHINE_NAME'):
+            return DockerVmType.DOCKER_ENGINE
+        else:
+            return DockerVmType.DOCKER_MACHINE
+
+
+def vm_name():
+    return os.getenv('DOCKER_MACHINE_NAME', DEFAULT_DOCKER_VM_NAME)
```
huntc, line 28: I now also have a `docker.py` in https://github.com/typesafehub/conductr-cli/pull/140 - Similar, but not the same constants. Mine don't assume "VM" to satisfy the native Docker use-case.

conductr_cli/docker.py
```diff
@@ -0,0 +1,28 @@
...
+def vm_type():
+    is_linux = platform.is_linux()
+    if is_linux:
+        return DockerVmType.DOCKER_ENGINE
+    else:
+        version = terminal.docker_version()
+        if version >= 1.12 and not os.getenv('DOCKER_MACHINE_NAME'):
+            return DockerVmType.DOCKER_ENGINE
+        else:
+            return DockerVmType.DOCKER_MACHINE
```
huntc, line 24: I think that this should work.

conductr_cli/docker_machine.py
```diff
@@ -1,9 +0,0 @@
-import os
-
-DEFAULT_DOCKER_MACHINE_NAME = 'default'
-DEFAULT_DOCKER_MACHINE_RAM_SIZE = '3072'
-DEFAULT_DOCKER_MACHINE_CPU_COUNT = '4'
```
huntc, line 5: I've retained these constants as they could indeed be different for docker-machine.

conductr_cli/sandbox_init.py
```diff
@@ -1,62 +1,73 @@
...
 import logging
 from subprocess import CalledProcessError
 
 
 def init(args):
     """`sandbox init` command"""
 
-    log = logging.getLogger(__name__)
-    vm_name = docker_machine.vm_name()
-    if is_docker_machine_installed():
```
huntc, line 11: I _think_ docker-machine can be installed on Linux also... so perhaps we wish to keep this logic around.

conductr_cli/docker.py
huntc, line 24: Actually, as mentioned below, I _think_ docker-machine can exist on Linux also...

markusjura:
This PR should also fix this logic around the `host.py`: https://github.com/typesafehub/conductr-cli/blob/master/conductr_cli/host.py#L11-L12

Currently, we use the IP address `127.0.0.1` if the `docker info` command exists. This command exist prior to docker 1.12. Also, it is possible that the user still uses `docker-machine` instead of `docker-engine` from docker 1.12 onwards.

markusjura:
The PR https://github.com/typesafehub/conductr-cli/pull/173 improved the Docker VM validation. It is also fully supports now Docker 1.12. Therefore, this PR is obsolete.

PR #140, opened 2016-06-29
Docker checking of memory and CPUs

huntc:
A general check is no provided in order to verify that native Docker, or Docker for Mac/Windows has enough resources. While we cannot change these values programmatically, we can at least report our findings.

The Docker Machine tests have also been tidied up in terms of their wording.


conductr_cli/terminal.py
```diff
@@ -1,6 +1,10 @@
 import subprocess
 
 
+def docker_info():
+    return subprocess.check_output(['docker', 'info']).strip()
```
markusjura, line 5: Does docker report this information also with 1.11 and below?

huntc, line 5: It appears to have been around for a while: https://docs.docker.com/v1.9/engine/reference/commandline/info/

conductr_cli/sandbox_init.py
```diff
@@ -1,62 +1,102 @@
...
+                log.warning('Docker has an insufficient no. of CPUs {} - please increase to a minimum of {} CPUs'
+                            .format(existing_cpu, docker.DEFAULT_DOCKER_CPU_COUNT))
+        else:
+            log.info('Docker already has sufficient RAM and CPUs')
+
+    except (AttributeError, CalledProcessError):
+        pass
+
+    docker_machine_vm_name = docker_machine.vm_name()
     if is_docker_machine_installed():
```
markusjura, line 32: We are only checking here if `docker-machine` is installed, but not if it is used. It might be the case that the user uses the `Docker.app` and therefore docker-engine, but still has docker-machine installed. In this case we would configure the docker-machine which would be unnecessary and misleading. This case is the default case when upgrading from Docker toolbox to Docker on Mac.

conductr_cli/sandbox_init.py
```diff
@@ -1,62 +1,102 @@
...
 from subprocess import CalledProcessError
 
 
 def init(args):
     """`sandbox init` command"""
 
     log = logging.getLogger(__name__)
-    vm_name = docker_machine.vm_name()
+
+    try:
```
markusjura, line 12: Why we don't do check which VM the user is using and based on that perform different init operations?

conductr_cli/sandbox_init.py
huntc, line 32: Separate issue?

conductr_cli/sandbox_init.py
huntc, line 12: Separate issue?

conductr_cli/terminal.py
markusjura, line 5: Nice. I've also tested what `docker info` reports when docker 1.10 and docker-machine is installed. In fact it then reports the VM and CPU configured by `docker-machine`. So this is a nice general way of finding out the resource settings of the VM.

markusjura:
LGTM

PR #141, opened 2016-07-01
Increased memory requirements

huntc:
As a result of further testing, we require 4GiB instead of 3Gib for running a ConductR cluster along with the full compliment of Chirper services, Cassandra etc.


markusjura:
LGTM

PR #152, opened 2016-08-30
Opens up the ES port (9200) when using monitoring so that ES can be p…

henrikengstrom:
…opulated from demo apps outside of ConductR.

@huntc are you okay with this? It would make our user experience a bit smoother (not having to add the `--port 9200` when starting sandbox with `--feature monitoring`).


huntc:
Sure thing!

PR #162, opened 2016-09-29
Cache directory migration

markusjura:
The PR https://github.com/typesafehub/conductr-cli/pull/161 creates now the default cache directory with the permissions 700 and validates if the specified cache directory has these secure permissions.

If a user has used a previous conductr-cli version than the default cache directory has been created with the permissions 755. This would lead to an error message when loading a bundle.

This PR is now checking if the specified cache directory during `conuct load` is the default cache directory, e.g. `~/.conductr/cache`. If that's the case and if the directory does not pass the secure permission check then we now set the permissions for the directory to 700.

Additionally, every file inside the cache directory is now created with the permissions 600 instead of 644.

### DIFF
--- a/conductr_cli/conduct_load.py
+++ b/conductr_cli/conduct_load.py
@@ -3,6 +3,7 @@ from pyhocon.exceptions import ConfigMissingException
 from conductr_cli import bundle_utils, conduct_url, validation
 from conductr_cli.exceptions import MalformedBundleError, InsecureFilePermissions
 from conductr_cli import resolver, bundle_installation
+from conductr_cli.constants import DEFAULT_BUNDLE_RESOLVE_CACHE_DIR
 from functools import partial
 
 import os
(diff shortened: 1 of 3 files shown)


conductr_cli/conduct_load.py
```diff
@@ -114,11 +115,17 @@ def get_payload(bundle_name, bundle_file, bundle_configuration):
...
-def validate_cache_dir_permissions(cache_dir):
+def validate_cache_dir_permissions(cache_dir, log):
     if os.path.exists(cache_dir):
         permissions = oct(stat.S_IMODE(os.lstat(cache_dir).st_mode))[-3:]
         if permissions[-2:] != '00':
-            raise InsecureFilePermissions('The cache directory {} has the permissions: {}'.format(cache_dir, permissions))
+            if cache_dir == DEFAULT_BUNDLE_RESOLVE_CACHE_DIR:
+                log.info('Cache directory {} has the permissions {}. Setting permissions to 700.'.format(cache_dir,
+                                                                                                         permissions))
+                os.chmod(cache_dir, 0o700)
```
huntc, line 125: Nice

huntc (approved):

PR #163, opened 2016-09-29
Support Akka SSE empty heartbeat lines

markusjura:
In 1.7.6 [akka-sse](https://github.com/hseeberger/akka-sse) changed the way how to sent heartbeat events. Instead of sending `data:` it is now sending an empty lines.

The akka-sse client in `conductr-cli` has been updated to support both the old and the new format.


markusjura:
Tested it both with ConductR 1.1.x branch (old akka-sse) and with the ConductR master branch (new akka-sse).

Issue #145, opened 2016-08-12
Remove running docker containers when starting the sandbox


markusjura:
The `sandbox run` command fails if docker containers exists that should get started:

```
$ sandbox run 1.1.8-rc.3 --feature visualization --nr-of-containers 3
WARNING: No swap limit support
Starting ConductR nodes..
Starting container cond-0 exposing 127.0.0.1:9999..
docker: Error response from daemon: Conflict. The name "/cond-0" is already in use by container 01170181f1051333739808948718800679d47c3776ed12762c6d58851516a125. You have to remove (or rename) that container to be able to reuse that name..
See 'docker run --help'.
```

Users that are not familiar with docker or doesn't know that ConductR uses docker don't know how to solve this problem. To make life easier I'd suggest:
- The sandbox run command by default checks if any `cond-` container is running before starting the new containers. If that is the case then it first kills these containers. Afterwards, it starts the specified containers (same as now).
- Additionally, we introduce the flag `--no-restart`. This flag prevents to kill the docker containers. This is useful when re-scaling an existing ConductR cluster.

PR #165, opened 2016-10-07
Ensure clean sandbox startup

markusjura:
So far the `sandbox run` command checked how many ConductR instances are already running. This made it possible to keep existing instances running and to scale down an existing cluster.

Also, to identify the currently running docker containers `docker ps` was used without the `--all` option.

Now, the `docker ps` command uses the `--all` option to also identify containers that have been crashed, e.g. due to a ConductR startup failure.

The "keep existing instances" and "scaling down" feature has been removed. Now, the `sandbox run` command stops all running containers before it starts a new ConductR cluster.

The reason behind this change is that users often had the problem that old `cond-` containers were still running when they were starting a new cluster, e.g. containers from a previous ConductR version or from an old cluster. The features "keep existing containers" and "scaling down" was hardly used. The new approach of stopping all containers before a new cluster is started ensures a clean state. This is similiar what we do in Lagom with the `install` command.

Fixes https://github.com/typesafehub/conductr-cli/issues/145.

### DIFF
--- a/conductr_cli/sandbox_run.py
+++ b/conductr_cli/sandbox_run.py
@@ -1,4 +1,4 @@
-from conductr_cli import conduct_url, sandbox_common, terminal, validation
+from conductr_cli import conduct_url, sandbox_common, terminal, validation, sandbox_stop
 from conductr_cli.constants import DEFAULT_PORT, DEFAULT_API_VERSION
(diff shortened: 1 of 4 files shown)


markusjura:
I'll add an option `--no-stopping`. If this is set then the cluster will not get stopped before it is getting started. By default, it will stop though.

huntc (approved):

PR #172, opened 2016-10-26
DCOS CLI Integration

huntc:
This commit provides the changes required in order for the "conduct" CLI to be integrated with the DC/OS CLI such that you can "dcos conduct ..." and the ConductR commands are channelled via DC/OS's admin gateway.


huntc:
Observational testing:

```
ChristohersMBP2:conductr-cli huntc$ dcos 
Command line utility for the Mesosphere Datacenter Operating
System (DC/OS). The Mesosphere DC/OS is a distributed operating
system built around Apache Mesos. This utility provides tools
for easy management of a DC/OS installation.

Available DC/OS commands:

    auth            Authenticate to DC/OS cluster
    conduct         Lightbend ConductR sub commands. Type 'dcos conduct' to see more.
    config          Manage the DC/OS configuration file
    help            Display help information about DC/OS
    marathon        Deploy and manage applications to DC/OS
    node            Administer and manage DC/OS cluster nodes
    package         Install and manage DC/OS software packages
    service         Manage DC/OS services
    task            Manage DC/OS tasks

Get detailed command description with 'dcos <command> --help'.
ChristohersMBP2:conductr-cli huntc$ 
ChristohersMBP2:conductr-cli huntc$ 
ChristohersMBP2:conductr-cli huntc$ dcos conduct info
ID       NAME        #REP  #STR  #RUN
c14bd62  visualizer     1     0     0
```

huntc:
Failing because of the DC/OS Python library unable to be found.

conductr_cli/conduct_main.py
```diff
@@ -5,15 +5,22 @@
...
 from pyhocon import ConfigFactory
+from urllib.parse import urlparse
 import os
 import sys
 
 
-def add_ip_and_port(sub_parser):
+def add_scheme_ip_port_and_base_path(sub_parser):
+    sub_parser.add_argument('--scheme',
+                            help='The optional ConductR scheme, defaults to `http`',
```
viktorklang, line 23: Shouldn't something as important as the orchestrator API default to https? (perhaps that's for the future?)

huntc: Agreed. TLS support for the control protocol is planned. Meanwhile our DC/OS integration here will provide TLS support.

conductr_cli/conduct_main.py
```diff
@@ -270,19 +299,47 @@ def get_custom_settings(args):
...
+        _args = _args[1:]
+        dcos_mode = True
+    else:
+        dcos_mode = False
     # Parse arguments
-    parser = build_parser()
+    parser = build_parser(dcos_mode)
     argcomplete.autocomplete(parser)
-    args = parser.parse_args(_args) if _args else parser.parse_args()
+    args = parser.parse_args(_args[1:])
```
markusjura, line 330: We are removing here always the first argument. Also in non DCOS mode. Does that makes sense?

conductr_cli/conduct_main.py
```diff
@@ -270,19 +299,47 @@ def get_custom_settings(args):
...
     if not vars(args).get('func'):
-        parser.print_help()
+        if not vars(args).get('dcos_info'):
```
markusjura: For readability reasons I'd would swap the if and else clause:

```
        if vars(args).get('dcos_info'):
            print('Lightbend ConductR sub commands. Type \'dcos conduct\' to see more.')
            exit(0)            
        else:
            parser.print_help()
```

conductr_cli/sandbox_run.py
```diff
@@ -203,7 +204,7 @@ def wait_for_conductr(args, current_retry, max_retries, interval):
         conduct_args = ConductArgs(args.vm_type)
         url = conduct_url.url('members', conduct_args)
         try:
-            requests.get(url, timeout=DEFAULT_HTTP_TIMEOUT, headers=conduct_url.request_headers(conduct_args))
+            conduct_request.get(False, ConductArgs.ip, url, timeout=DEFAULT_HTTP_TIMEOUT)
```
markusjura: Always specify the named argument when passing in a boolean:

```
conduct_request.get(dcos_mode=False, ConductArgs.ip, url, timeout=DEFAULT_HTTP_TIMEOUT)
```

conductr_cli/test/conduct_load_test_base.py
```diff
@@ -64,23 +63,20 @@ def base_test_success(self):
...
             logging_setup.configure_logging(input_args, stdout)
             result = conduct_load.load(input_args)
             self.assertTrue(result)
 
         open_mock.assert_called_with(self.bundle_file, 'rb')
         resolve_bundle_mock.assert_called_with(self.custom_settings, self.bundle_resolve_cache_dir, self.bundle_file)
-        request_headers_mock.assert_called_with(input_args)
         http_method.assert_called_with(self.default_url, files=self.default_files, timeout=LOAD_HTTP_TIMEOUT,
-                                       headers=self.mock_headers)
+                                       headers={'Host': '127.0.0.1'})
```
markusjura, line 74: Why is this change necessary?

markusjura (changes requested):

markusjura:
> conduct init needs to create a soft link to the `conduct.py` file from `~/.dcos/subcommands/conductr/env/bin/dcos-conduct`

What do you mean by that? We only have a `sandbox init` command but this should not be involved in DCOS mode.

huntc:
> What do you mean by that? We only have a sandbox init command but this should not be involved in DCOS mode.

In order to make the DC/OS CLI aware of the ConductR CLI, you have to tell it. Ordinarily, this is done via `dcos install --cli <package>`, but we don't have a universal package available (yet). So, meanwhile then, `conduct init` will automatically set things up.

I just realised though that we have no `conduct init`, so we'll go with your IM suggestion of a new `conduct setup-dcos`.

conductr_cli/conduct_main.py
huntc: Yes. This is what the previous code was actually doing under the hood.

conductr_cli/conduct_main.py
huntc: Why would that be more readable?

conductr_cli/sandbox_run.py
huntc: ok, but I do then have to specify the other arg names.

conductr_cli/test/conduct_load_test_base.py
huntc: We were previously checking the host header being set up via another mock. The other mock was no longer available due to its function having been replaced. The code should also be much simpler now.

conductr_cli/conduct_main.py
markusjura: We are saving a "not". In a if / else statement you should never start with the negative, but always with the positive case.

conductr_cli/sandbox_run.py
markusjura: Your decision. Not critical.

conductr_cli/test/conduct_load_test_base.py
markusjura: 👍

markusjura (approved):

conductr_cli/conduct_main.py
markusjura: 👍

conductr_cli/conduct_main.py
huntc: There's also an argument to show the happy path, but I'll go with it.

huntc:
Waiting on a new DCOS lib release. PR tests will fail here until the following PR is merged and there's a new release of the DCOS lib: https://github.com/dcos/dcos-cli/pull/823

markusjura:
Manual integration test with `http://dcos-master.lightbend.com` was successful.

PR #178, opened 2016-11-01
Reduce the Docker Machine check

huntc:
The existing check for Docker Machine was perhaps too tight. It is a valid scenario to have a native Docker installed, along with DOCKER_HOST and others e.g. in the future when connecting the Docker API with ConductR.

The only check required for Docker Machine is the DOCKER_MACHINE_NAME environment variable.

### DIFF
--- a/conductr_cli/docker.py
+++ b/conductr_cli/docker.py
@@ -16,12 +16,9 @@ class DockerVmType:
 def vm_type():
     # TODO: Add Docker native check for Windows 10 Professional and Enterprise Edition
     docker_machine_name_env = os.getenv('DOCKER_MACHINE_NAME')
-    docker_host_env = os.getenv('DOCKER_HOST')
-    docker_tls_verify_env = os.getenv('DOCKER_TLS_VERIFY')
-    docker_cert_path_env = os.getenv('DOCKER_CERT_PATH')
 
     def docker_machine_envs_set():
-        if docker_machine_name_env or docker_host_env or docker_tls_verify_env or docker_cert_path_env:
+        if docker_machine_name_env:
             return True
         else:
             return False


markusjura:
You mean the `DOCKER_MACHINE_NAME` env ;-)

conductr_cli/validation.py
```diff
@@ -219,11 +219,7 @@ def handler(*args, **kwargs):
             log = get_logger_for_func(func)
             log.error('Docker native is installed and Docker machine environment variables are set.')
             log.error('It is uncertain which Docker VM should be used.')
-            log.error('If Docker native should be used please unset the Docker machine environment variables:')
-            log.error('  DOCKER_CERT_PATH')
-            log.error('  DOCKER_HOST')
-            log.error('  DOCKER_MACHINE_NAME')
-            log.error('  DOCKER_TLS_VERIFY')
+            log.error('If Docker native should be used please unset the DOCKER_MACHINE_NAME environment variable.')
```
markusjura: This advice can lead to misfunctioning Docker.

**Example**
- The user is setting the Docker machine VM variables with:

```
eval "$(docker-machine env default)"
```
- Now the user has also started Docker native. As a result, he is receiving the error message above
- Unsetting only `DOCKER_MACHINE_NAME` will not work. The native docker daemon will not be able to connect to the native docker host because `DOCKER_HOST` points to the VirtualBox VM Docker host. In this case, the user need to unset all 4 environment variables.

markusjura (changes requested):

conductr_cli/validation.py
markusjura: If `DOCKER_MACHINE_NAME` is set then we know that Docker machine is enabled and that the user needs to unset all 4 envs. So the solution here is to keep the old error message, but still only do the check on `DOCKER_MACHINE_NAME`.

markusjura (approved):

PR #193, opened 2016-11-14
Change memory unit from MiB to GiB

markusjura:
The Docker native command prints out the dedicated memory in GiB. We have displayed it as MiB so far.

The Docker machine command prints out the dedicated memory in MB. We have displayed it as MiB.

The PR is changing MiB to GiB for the Docker native output, and MiB to MB for the Docker machine output.


huntc (commented):
Lgtm

PR #201, opened 2016-11-15
Use requests library >= 2.6.0 and < 2.12.0

fsat:
This is because the requests library `2.12.0` ships with the version of urllib which is not backward compatible shipped with prior release of requests `2.11.1` and below.


huntc (approved):
I'm still having a local problem with `tox`, but if this solves the issue let's take it for a spin.

PR #210, opened 2016-11-17
Perform VM checks for each sandbox command

markusjura:
The VM validation of the `sandbox init` command in now performed in each sandbox command, making the init command obsolete.

This approach is much simpler for the user and less errorprone. It can not happen anymore that the user has not executed the `sandbox init` command and therefore the `sandbox run` command failed, e.g. due to insufficient VM resources. In case the VM is not setup or configured correctly, the sandbox run and stop command now automatically try to resolve the issue - as the sandbox init command did it in the past.

The sandbox init command has been removed.

This PR also adds additional unit tests around VM validation.

Tested all different VM cases manually on macOS.


huntc:
I'm ok with removing the init command but it must be removed from our website getting started material also.

markusjura:
Yes, let's do it once the PR is merged and the new conductr-cli is released.

huntc (approved):

PR #226, opened 2016-12-01
Stronger file check

### LINKED ISSUE #225: Bundle caching broken?
The recent changes to the sandbox appear to have affected the caching ability of the load command. It appears as though bundles are downloaded each time I run the sandbox.

huntc:
This commit strengthens the check to determine whether a bundle exists on the file system by ensuring that it has a ".zip" extension. Prior to doing this the `load` sub command could yield a false-positive and cause the cache check to be avoided when a directory of the same name as the bundle existed. This problematic scenario would occur when using the load sub command from within the ConductR project itself given that there are directories for eslite, visualizer etc. CI would then also keep resolving bundles from bintray despite them possibly being cached from other tests.

Tested from within the ConductR project directory and outside of it. Also re-tested by uploading a file from the file system.

Fixes #225

### DIFF
--- a/conductr_cli/resolvers/bintray_resolver.py
+++ b/conductr_cli/resolvers/bintray_resolver.py
@@ -31,8 +31,7 @@ def resolve_bundle(cache_dir, uri):
 
 def load_bundle_from_cache(cache_dir, uri):
     # When the supplied uri points to a local file, don't load from cache so file can be used as is.
-    parsed = urlparse(uri, scheme='file')
-    if parsed.scheme == 'file' and os.path.exists(uri):
+    if is_local_file(uri):
         return False, None, None
     else:
         log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


conductr_cli/resolvers/bintray_resolver.py
```diff
@@ -32,7 +32,7 @@ def resolve_bundle(cache_dir, uri):
 def load_bundle_from_cache(cache_dir, uri):
     # When the supplied uri points to a local file, don't load from cache so file can be used as is.
     parsed = urlparse(uri, scheme='file')
-    if parsed.scheme == 'file' and os.path.exists(uri):
+    if parsed.scheme == 'file' and parsed.path.endswith('.zip') and os.path.exists(uri):
```
fsat: This also needs to be applied to `load_bundle_configuration_from_cache`:

https://github.com/typesafehub/conductr-cli/pull/226/files#diff-fd50a12e0d19d40e43b9178ed70df60fR68

fsat (changes requested):
Can we have additional test scenarios in `test_bintray_resolver` to cover both loading of bundle and bundle configuration  where path exists but not ending with `.zip` please?

conductr_cli/resolvers/bintray_resolver.py
huntc: Nice find - I shall make it so.

fsat (approved):

markusjura:
@huntc Does this also fix the issue that the bundle is loaded from bintray instead of the cache directory?

Issue #223, opened 2016-11-29
Waiting for bundle replication before running bundle


markusjura:
If the sandbox is started on multiple containers it is possible that feature can not be scaled. This is because the `conduct load` command is waiting that a bundle has been loaded, but it does not wait that the bundle has been replicated to the nodes. A subsequent `conduct run` can fail if it hits a node where the bundle has not been replicated yet.

**Sample output**
```
|------------------------------------------------|
| Starting visualization feature                 |
|------------------------------------------------|
Deploying bundle visualizer..
Retrieving bundle..
Loading bundle from cache typesafe/bundle/visualizer
Bintray credentials loaded from /Users/mj/.bintray/.credentials
Retrieving file:///Users/mj/workspace/conductr-cli/visualizer
Resolving bundle typesafe/bundle/visualizer
Bintray credentials loaded from /Users/mj/.bintray/.credentials
Retrieving https://dl.bintray.com/typesafe/bundle/typesafe/visualizer/v2-7fe30fb02ff5df19b09fa33cfe03c5cfa4eb7fb4ff8d7422b9662074eca8b333/visualizer-v2-7fe30fb02ff5df19b09fa33cfe03c5cfa4eb7fb4ff8d7422b9662074eca8b333.zip
[#################################################] 100%
Loading bundle to ConductR..
[#################################################] 100%
Bundle 7fe30fb02ff5df19b09fa33cfe03c5cf waiting to be installed
Bundle 7fe30fb02ff5df19b09fa33cfe03c5cf installed
Bundle loaded.
Error: 404 Not Found
Error: No bundle found by the specified Bundle ID/name: '7fe30fb02ff5df19b09fa33cfe03c5cf'
```

huntc:
This problem is now occurring quite often for me. I would have thought that if the `run` was sent to the same node as the `load` then all would be well, particularly given that we're waiting on the load event to complete perform moving on to the run.

We need to fix it though as it has become unreliable. Perhaps we should retry at least 3 times or something.

typesafe-tools:
![](https://github.trello.services/images/mini-trello-icon.png) [The new CLI features loading fails regularly - we need to address this.](https://trello.com/c/YC1nnrJV/439-the-new-cli-features-loading-fails-regularly-we-need-to-address-this)

fsat:
[...]

PR #232, opened 2016-12-09
WIP - DON'T MERGE - Conduct load now waits for number of required replicas to be present

fsat:
The number of required replicas is derived from the members endpoint, counting the number of members which has the `replicator` role.


fsat:
Fixes #223

fsat:
Closing this - we'll try the solution from within ConductR itself.

PR #244, opened 2017-01-13
Stop sandbox JVM mode

markusjura:
Implementation of the `sandbox stop` command for the ConductR JVM binaries.

The `sandbox stop` command is stopping both the docker containers and the JVM processes. If no containers or processes are running then no output is produced. If either the JVM or Docker stop function cannot stop the processes or containers respectively then an error is reported and the stop function returns False.

This PR also introduces additional unit tests for the `sandbox_stop_docker` module to validate the scenarios:
- No containers running
- Cannot stop containers

The PR has been tested both with ConductR 1.1.x and ConductR 2.0.0-rc.2

### DIFF
--- a/conductr_cli/sandbox_common.py
+++ b/conductr_cli/sandbox_common.py
@@ -1,12 +1,8 @@
 from conductr_cli import terminal
+from conductr_cli.resolvers.bintray_resolver import BINTRAY_CONDUCTR_CORE_PACKAGE_NAME, \
+    BINTRAY_CONDUCTR_AGENT_PACKAGE_NAME
 import os
-from enum import Enum
 from subprocess import CalledProcessError
-
-
-class ConductrComponent(Enum):
-    CORE = 1
-    AGENT = 2
 
 
 CONDUCTR_NAME_PREFIX = 'cond-'
@@ -14,7 +10,20 @@ CONDUCTR_PORTS = {9004,  # ConductR internal akka remoting
                   9005,  # ConductR controlServer
                   9006}  # ConductR bundleStreamServer
 CONDUCTR_DEV_IMAGE = 'typesafe-docker-registry-for-subscribers-only.bintray.io/conductr/conductr'
-LATEST_CONDUCTR_VERSION = '1.1.11'
+
+
+def resolve_conductr_info(image_dir):
+    core_info = {
+        'type': 'core',
+        'extraction_dir': '{}/core'.format(image_dir),
+        'bintray_package_name': BINTRAY_CONDUCTR_CORE_PACKAGE_NAME
+    }
+    agent_info = {
+        'type': 'agent',
+        'extraction_dir': '{}/agent'.format(image_dir),
+        'bintray_package_name': BINTRAY_CONDUCTR_AGENT_PACKAGE_NAME
+    }
+    return core_info, agent_info
 
 
 def resolve_running_docker_containers():
--- a/conductr_cli/sandbox_run_jvm.py
+++ b/conductr_cli/sandbox_run_jvm.py
@@ -1,12 +1,11 @@
-from conductr_cli import conduct_main, host, sandbox_stop
+from conductr_cli import conduct_main, host, sandbox_stop, sandbox_common
 from conductr_cli.constants import DEFAULT_SCHEME, DEFAULT_PORT, DEFAULT_BASE_PATH, DEFAULT_API_VERSION
-from conductr_cli.sandbox_common import ConductrComponent
 from conductr_cli.exceptions import BindAddressNotFoundError, BintrayUnreachableError, InstanceCountError, \
     SandboxImageNotFoundError, JavaCallError, JavaUnsupportedVendorError, JavaUnsupportedVersionError, \
     JavaVersionParseError
 from conductr_cli.resolvers import bintray_resolver
 from conductr_cli.resolvers.bintray_resolver import BINTRAY_DOWNLOAD_REALM, BINTRAY_LIGHTBEND_ORG, \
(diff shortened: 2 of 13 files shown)


conductr_cli/sandbox_common.py
```diff
@@ -14,6 +10,23 @@ class ConductrComponent(Enum):
                   9006}  # ConductR bundleStreamServer
 CONDUCTR_DEV_IMAGE = 'typesafe-docker-registry-for-subscribers-only.bintray.io/conductr/conductr'
 LATEST_CONDUCTR_VERSION = '1.1.11'
+RUNNING_CONDUCTR_VERSION_FILE_PATH = '{}/RUNNING_CONDUCTR_VERSION'.format(DEFAULT_CLI_SETTINGS_DIR)
+
+
+def resolve_conductr_binrary_info(image_dir, image_version):
```
fsat: There's a spelling error - perhaps `resolve_conductr_bintray_info`?

conductr_cli/sandbox_stop_jvm.py
```diff
@@ -6,86 +15,71 @@ def stop(args):
...
-    of the sandbox image.
-    :param run_dir: directory of where ConductR agent is running from.
-    :return: the list of the ConductR agent pids.
-    """
-    pass
-
-
-def kill_processes(pids):
+    log = logging.getLogger(__name__)
+    core_info, agent_info = sandbox_common.resolve_conductr_binrary_info(args.image_dir, args.image_version)
```
fsat: The method should be `resolve_conductr_bintray_info`

conductr_cli/sandbox_run_jvm.py
```diff
@@ -53,6 +52,7 @@ def run(args, features):
                      This is only relevant for Docker based sandbox since the features decides what port to expose
     :return: SandboxRunResult
     """
+    sandbox_common.LATEST_VERSION=args.image_version
```
fsat: There's no constant called `LATEST_VERSION` declared in `sandbox_common`, but there's `LATEST_CONDUCTR_VERSION`.

I don't think we should be setting a constant like this (although `LATEST_VERSION` is a variable, but by convention that's it's all uppercase, it should be treated as a constant).

I don't think `LATEST_VERSION` is used anywhere, we shouldn't need to set them.

conductr_cli/sandbox_common.py
```diff
@@ -14,6 +10,23 @@ class ConductrComponent(Enum):
                   9006}  # ConductR bundleStreamServer
 CONDUCTR_DEV_IMAGE = 'typesafe-docker-registry-for-subscribers-only.bintray.io/conductr/conductr'
 LATEST_CONDUCTR_VERSION = '1.1.11'
```
fsat: We should move this into `sandbox_run_docker`, or perhaps remove it altogether since it's not used anywhere but in test.

I'm happy to do this in my own PR.

conductr_cli/sandbox_stop.py
```diff
@@ -1,8 +1,16 @@
...
 
-    sandbox_stop_docker.stop(args)
+    if not hasattr(args, 'image_version'):
+        with open(RUNNING_CONDUCTR_VERSION_FILE_PATH, 'r') as file:
+            args.image_version = file.read()
+
+    if args.image_version == 1:
+        sandbox_stop_docker.stop(args)
+    elif args.image_version != 1:
+        sandbox_stop_jvm.stop(args)
```
fsat: I think we should call stop for both docker and jvm based sandbox, since I think the user would expect both sandbox to stop if they are both running.

fsat (changes requested):

conductr_cli/sandbox_stop.py
markusjura: If the `sandbox_stop_docker.stop` function is called and docker is not running then we currently throw this error:

```
Error: Encountered unexpected error.
Error: Reason: CalledProcessError Command '['docker', 'ps', '--all', '--quiet', '--filter', 'name=cond-']' returned non-zero exit status 1.
Error: Further information of the error can be found in the error log file: /Users/mj/.conductr/errors.log
```

conductr_cli/sandbox_run_jvm.py
markusjura: This was an oversight of something that I've tried. Removed it.

conductr_cli/sandbox_stop_jvm.py
markusjura: Renamed it to `resolve_conductr_info` because now it does not contain only binary related information.

conductr_cli/sandbox_common.py
markusjura: I'll remove it.

markusjura:
This PR also moves the constant `LATEST_CONDUCTR_VERSION` into `conductr_cli/test/data/test_constants.py` because this constants is only used during testing.

conductr_cli/sandbox_run_jvm.py
```diff
@@ -338,13 +308,12 @@ def extract_binary(path, binary_info):
         os.rmdir(extraction_subdir)
         return extraction_dir
 
-    core_binary_info = resolve_binary_info(ConductrComponent.CORE)
-    agent_binary_info = resolve_binary_info(ConductrComponent.AGENT)
+    core_info, agent_info = sandbox_common.resolve_conductr_info(image_dir)
 
     core_binary_path, agent_binary_path = resolve_binaries()
```
fsat, line 313: Just a suggestion, feel free to skip if you feel this is unnecessary.

I think it will be much more readable if `core_info` and `agent_info` is passed into `resolve_binaries`, i.e.

```
     core_binary_path, agent_binary_path = resolve_binaries(core_info, agent_info)
```

Otherwise, the `resolve_binaries` as a backwardly declared reference to `core_info` and `agent_info`. Plus `core_info` and `agent_info` are passed into the `extract_binary` in the next two lines, so it's good to be consistent.

fsat (approved):

PR #249, opened 2017-01-16
Docker validation for logging feature

markusjura:
In ConductR 2.x (JVM mode) the conductr-kibana bundle is a Docker based bundle. Therefore, when the logging feature is specified, it is checked if Docker is installed and correctly configured.

The logging feature uses the same Docker validation and error output as the sandbox run feature when using the Docker related ConductR binary.

An additional Docker validation for the monitoring feature is not necessary because it depends on the logging feature, i.e. the logging feature will be started first and will fail if Docker is not setup correctly.

**Docker validation output**

```
$ sandbox run 2.0.0-rc.2 --feature logging
Extracting ConductR core to /Users/mj/.conductr/images/core
Extracting ConductR agent to /Users/mj/.conductr/images/agent
Starting ConductR core instance 0 on 192.168.10.1..
Starting ConductR agent instance 0 on 192.168.10.1..
Waiting for ConductR to start.......
…
| Starting logging feature based on              |
| elasticsearch and kibana                       |
conductr-kibana bundle is packaged as a Docker image. Checking Docker requirements..
Error: Docker native is installed but not running.
Error: Please start Docker with one of the Docker flavors based on your OS:
Error:   Linux:   Docker service
Error:   MacOS:   Docker for Mac
Error:   Windows: Docker for Windows
Error: A successful Docker startup can be verified with: docker info
```


conductr_cli/docker.py
```diff
@@ -48,3 +50,106 @@ def cpu_check(info):
...
+                    log.warning('Docker has an insufficient no. of CPUs {} - please increase to a minimum of {} CPUs'
+                                .format(existing_cpu, DEFAULT_DOCKER_CPU_COUNT))
+        except (AttributeError, CalledProcessError):
+            log.error('Docker native is installed but not running.')
+            log.error('Please start Docker with one of the Docker flavors based on your OS:')
+            log.error('  Linux:   Docker service')
+            log.error('  MacOS:   Docker for Mac')
+            log.error('  Windows: Docker for Windows')
+            log.error('A successful Docker startup can be verified with: docker info')
+            exit(1)
```
fsat, line 79: I'm wondering if it's possible to `raise` an exception and have the actual `sandbox_run.run` handle this exception with `@validation.handle_vbox_manage_not_found_error` and `@validation.handle_docker_vm_validation_error` instead.

Also have the exception store the appropriate error message, and have these error messages logged in the validation handler `@validation.handle_docker_vm_validation_error` instead.

The reason for this is I might need to reuse this validation during HAProxy start/stop, and having the method invoking `exit` will mean that I can't use this method when I want to stop processing and continue.

When stopping HAProxy, if there's no docker installed, it's safe to assume that HAProxy docker container is not running, hence no error message need to be displayed. If this method raises an exception instead, I can call this method with `try/catch` block.

fsat (changes requested):

conductr_cli/docker.py
fsat: After further discussion, we decided not to pursue this. The HAProxy stop can trap docker related error and then continue silently.

fsat (approved):

PR #260, opened 2017-01-19
Offline mode

markusjura:
**Issue description**
In the past the `conduct load` command always retrieved the version of a specified bundle name from Bintray. With the resolved version it checked if the bundle is part of the cache directory. As a result, when specifying a bundle name and not a file URI, it was not possible to load a bundle without an internet connection. This especially turned out to be an issue when starting the sandbox with certain features, because the feature logic uses bundle names and not file URIs to load bundle to ConductR.

**Solution description**
This PR introduces a `—ofline` flag for the `conduct load` and `sandbox run` command to enable the offline mode. It is also possible to set the environment variable `CONDUCTR_OFFLINE_MODE=true`. Then the `—offline` flag is not necessary.

If the offline mode is enabled, then:
- The bundle is never retrieved from bintray resolver
- The `load_bundle_from_cache` and `load_bundle_configuration_from_cache` methods are resolving the last modified file for a given bundle name, e.g. visualizer. This logic is only executed if the `conduct load` command specifies a bundle name, e.g. `conduct load visualizer`, not if a file URI is specified, e.g. `conduct load /dir/my-bundle.zip`.

As a result, the sandbox and the conduct load command can be used without an internet connection in case the bundles are part of the cache directory.

### DIFF
--- a/conductr_cli/resolver.py
+++ b/conductr_cli/resolver.py
@@ -6,13 +6,14 @@ import logging
 
 # Try to resolve from local file system before we attempting resolution using bintray
 DEFAULT_RESOLVERS = [uri_resolver, bintray_resolver]
+OFFLINE_RESOLVERS_EXCLUSION = [bintray_resolver]
 
 
-def resolve_bundle(custom_settings, cache_dir, uri):
-    all_resolvers = resolver_chain(custom_settings)
+def resolve_bundle(custom_settings, cache_dir, uri, offline_mode=False):
+    all_resolvers = resolver_chain(custom_settings, offline_mode)
 
     for resolver in all_resolvers:
-        is_cached, bundle_file_name, cached_bundle = resolver.load_bundle_from_cache(cache_dir, uri)
+        is_cached, bundle_file_name, cached_bundle = resolver.load_bundle_from_cache(cache_dir, uri, offline_mode)
         if is_cached:
             return bundle_file_name, cached_bundle
 
@@ -24,12 +25,12 @@ def resolve_bundle(custom_settings, cache_dir, uri):
     raise BundleResolutionError('Unable to resolve bundle using {}'.format(uri))
 
 
-def resolve_bundle_configuration(custom_settings, cache_dir, uri):
+def resolve_bundle_configuration(custom_settings, cache_dir, uri, offline_mode=False):
     all_resolvers = resolver_chain(custom_settings)
 
     for resolver in all_resolvers:
         is_cached, bundle_configuration_file_name, cached_bundle = \
-            resolver.load_bundle_configuration_from_cache(cache_dir, uri)
+            resolver.load_bundle_configuration_from_cache(cache_dir, uri, offline_mode)
         if is_cached:
             return bundle_configuration_file_name, cached_bundle
 
@@ -64,7 +65,7 @@ def continuous_delivery_uri(custom_settings, resolved_version):
     raise ContinuousDeliveryError('Unable to form Continuous Delivery uri using {}'.format(resolved_version))
 
 
-def resolver_chain(custom_settings):
+def resolver_chain(custom_settings, offline_mode):
     log = logging.getLogger(__name__)
     if custom_settings is not None and 'resolvers' in custom_settings:
         resolver_names = custom_settings.get_list('resolvers')
@@ -72,5 +73,8 @@ def resolver_chain(custom_settings):
             log.info('Using custom bundle resolver chain {}'.format(resolver_names))
             custom_resolver_chain = [importlib.import_module(resolver_name) for resolver_name in resolver_names]
             return custom_resolver_chain
-
-    return DEFAULT_RESOLVERS
+    else:
+        if offline_mode:
+            return list(set(DEFAULT_RESOLVERS) - set(OFFLINE_RESOLVERS_EXCLUSION))
+        else:
+            return DEFAULT_RESOLVERS
--- a/conductr_cli/resolvers/uri_resolver.py
+++ b/conductr_cli/resolvers/uri_resolver.py
@@ -4,6 +4,7 @@ from urllib.error import URLError
 from pathlib import Path
 from conductr_cli import screen_utils
 import os
+import glob
 import logging
 import shutil
 import urllib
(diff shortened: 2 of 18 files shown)


markusjura:
## Manual tests

**Offline mode: conduct load**

```
$ conduct load --offline visualizer
Retrieving bundle..
Retrieving from cache /Users/mj/.conductr/cache/visualizer-v1.1-1c762907e71df7cd310b889e85424dcf4213d24d44962f8b8fe1668d9723bfe0.zip
Loading bundle to ConductR..
[#################################################] 100%
Bundle 1c762907e71df7cd310b889e85424dcf waiting to be installed
Bundle 1c762907e71df7cd310b889e85424dcf installed
Bundle loaded.
Start bundle with: conduct run 1c76290
Unload bundle with: conduct unload 1c76290
Print ConductR info with: conduct info
```

**Offline mode: sandbox run**

```
$ sandbox run 1.1.12 --feature monitoring --offline
Starting container cond-0 exposing 127.0.0.1:3000, 127.0.0.1:5601, 127.0.0.1:9200..
c01780f997d0fa1a8384a21f8e767e67771dcf5568e9972b38e077f96dd38547
Waiting for ConductR to start..
| Starting monitoring feature                    |
Deploying bundle cinnamon-grafana..
Retrieving bundle..
Retrieving from cache /Users/mj/.conductr/cache/cinnamon-grafana-v2-3c8bc952abe1b49dafcf1793ba3a3057b52f0257d385a3ebe20359b9c73e5517.zip
Loading bundle to ConductR..
[#################################################] 100%
Bundle 3c8bc952abe1b49dafcf1793ba3a3057 waiting to be installed
Bundle 3c8bc952abe1b49dafcf1793ba3a3057 installed
Bundle loaded.
Bundle run request sent.
Bundle 3c8bc952abe1b49dafcf1793ba3a3057 waiting to reach expected scale 1
Bundle 3c8bc952abe1b49dafcf1793ba3a3057 has scale 0, expected 1..........
Bundle 3c8bc952abe1b49dafcf1793ba3a3057 expected scale 1 is met
| Summary                                        |
ConductR has been started
Check resource consumption of Docker container that run the ConductR node with:
  docker stats cond-0
Check current bundle status with:
  conduct info
```

**Offline mode set by environment variable**

```
$ export CONDUCTR_OFFLINE_MODE=true
$ conduct load visualizer
Retrieving bundle..
Retrieving from cache /Users/mj/.conductr/cache/visualizer-v1.1-1c762907e71df7cd310b889e85424dcf4213d24d44962f8b8fe1668d9723bfe0.zip
Loading bundle to ConductR..
[#################################################] 100%
Bundle 1c762907e71df7cd310b889e85424dcf waiting to be installed
Bundle 1c762907e71df7cd310b889e85424dcf installed
Bundle loaded.
Start bundle with: conduct run 1c76290
Unload bundle with: conduct unload 1c76290
Print ConductR info with: conduct info
```

conductr_cli/resolver.py
```diff
@@ -64,13 +65,16 @@ def continuous_delivery_uri(custom_settings, resolved_version):
...
         resolver_names = custom_settings.get_list('resolvers')
         if resolver_names:
             log.info('Using custom bundle resolver chain {}'.format(resolver_names))
             custom_resolver_chain = [importlib.import_module(resolver_name) for resolver_name in resolver_names]
             return custom_resolver_chain
-
-    return DEFAULT_RESOLVERS
+    else:
+        if offline_mode:
+            return list(set(DEFAULT_RESOLVERS) - set(OFFLINE_RESOLVERS_EXCLUSION))
```
fsat, line 78: Would you consider having a separate set of offline resolver, i.e.

```
DEFAULT_RESOLVERS = [uri_resolver, bintray_resolver]
OFFLINE_RESOLVERS = [offline_resolver]
```

This would mean then you can do something like:

```
return OFFLINE_RESOLVERS if offline_mode else DEFAULT_RESOLVERS
```

I think having a separate set of offline resolvers will be a cleaner approach, wdyt?

conductr_cli/resolvers/uri_resolver.py
```diff
@@ -37,10 +38,31 @@ def resolve_file(cache_dir, uri, auth=None):
...
+    local file can be loaded directly.
+    :param cache_dir: the cache directory
+    :param uri: the bundle uri. Can be either a bundle name, e.g. 'visualizer, an http or file uri
+    :param offline_mode: the offline mode flag
+    :return: a tuple of (is_cached, bundle_name, bundle_uri)
+    """
     # When the supplied uri is a local filesystem, don't load from cache so file can be used as is
     parsed = urlparse(uri, scheme='file')
-    if parsed.scheme == 'file':
+    if offline_mode and is_bundle_name(uri):
```
fsat, line 55: Would you consider having a separate resolver called `offline_resolver`?

Having a separate `offline_resolver` means the offline logic can be contained within the `offline_resolver`, and the `uri_resolver` implementation will be purely uri-related.

I think this would be a cleaner approach - wdyt?

If you do agree with this approach, I think optionally the `offline_mode` input argument can be removed from the `load_bundle_from_cache ` method.

fsat (changes requested):

fsat:
[ConductR CLI - The CLI now requires an offline mode so that the sandbox can work while untethered from the internet.](https://trello.com/c/NnOAYNt2/441-conductr-cli-the-cli-now-requires-an-offline-mode-so-that-the-sandbox-can-work-while-untethered-from-the-internet)

conductr_cli/resolver.py
huntc: An offline resolver sounds like a nice approach

huntc: Unsure why we must return a different list here. Each resolver should handle being offline (which I think they do). It feels leaky to assume which resolvers are for offline here.

conductr_cli/resolvers/uri_resolver.py
huntc: Actually I think each resolver should handle what it means to be offline.

huntc (changes requested):

fsat:
Closing in lieu of #262 - thanks @markusjura !

PR #266, opened 2017-01-24
Friendlier message when no bind addresses

huntc:
As this error always appears for first time users, we should present it in a friendlier fashion. This commit attempts to do that.

### DIFF
--- a/conductr_cli/host.py
+++ b/conductr_cli/host.py
@@ -1,6 +1,5 @@
 import os
 from conductr_cli.constants import DEFAULT_PORT, DEFAULT_SANDBOX_ADDR_RANGE
-from conductr_cli.screen_utils import headline
 import ipaddress
 import socket
 import platform
@@ -81,10 +80,9 @@ def is_listening(ip_addr, port):
 
 
 def addr_alias_setup_instructions(addrs, subnet_mask):
-    def info_text(commands):
-        return headline(
-            'Whoops. Network address aliases are required so that the sandbox can operate as a cluster of machines') + \
-            '\n\nPlease run the following and then try your command again:'
+    info_text = 'Whoops. Network address aliases are required ' \
+                'so that the sandbox can operate as a cluster of machines.' + \
+                '\n\nPlease run the following and then try your command again:'
 
     if_name = loopback_device_name()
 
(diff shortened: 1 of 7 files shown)


huntc:
Sample output:

```
ChristohersMBP2:conductr-cli huntc$ sandbox run 2.0.0-rc.3 -n1:3 -f visualization -f logging -f monitoring
| Whoops. Network address aliases are required   |
| so that the sandbox can operate as a cluster   |
| of machines                                    |

Please run the following and then try your command again:

sudo ifconfig lo0 alias 192.168.10.1 255.255.255.0
sudo ifconfig lo0 alias 192.168.10.2 255.255.255.0
sudo ifconfig lo0 alias 192.168.10.3 255.255.255.0
```

conductr_cli/host.py
```diff
@@ -81,10 +82,9 @@ def is_listening(ip_addr, port):
 
 def addr_alias_setup_instructions(addrs, subnet_mask):
     def info_text(commands):
-        if len(commands) > 1:
-            return 'Run the following commands to create the address aliases on your machine:'
-        else:
-            return 'Run the following command to create an address alias on your machine:'
+        return headline(
```
markusjura: I would not use a headline here. Headlines should we used for certain sub sections to start ConductR, such as `Starting HAProxy`.

markusjura (changes requested):

markusjura:
Unit tests are failing due to not updated tests.

conductr_cli/host.py
huntc: Ok but I think it needs to standout somehow

huntc:
I had problems running all tests due to the docker low on mem error throwing things out. I was being lazy. Apologies.

conductr_cli/host.py
markusjura: I am wondering if we should add the address aliases automatically after the user entered the sudo password. When prompting for the password, we could explain what we are doing. This way we wouldn’t need to highlight the message in a particular fashion.

What do you think?

markusjura:
Fixed the unit tests. Also, the message `Whoops. Network address aliases are required so that the sandbox can operate as a cluster of machines.` is not marked as a headline anymore.

**New output**

```
Whoops. Network address aliases are required so that the sandbox can operate as a cluster of machines.

Please run the following and then try your command again:

sudo ifconfig lo0 alias 192.168.10.1 255.255.255.0
```

markusjura (approved):

PR #288, opened 2017-01-31
ReadMe update for building sandbox native command

fsat:
Ensure `psutil` is included when building sandbox native executable through `--hidden-import psutil` option.

If this option is not specified, `psutil` will not be included, causing error when sandbox native executable is invoked.


huntc (approved):

huntc:
Could we not provide a spec file?

PR #300, opened 2017-02-06
Delete files under old cache directory

markusjura:
Each `conduct load` call will delete the files under the `~/.cache` directory. It will **not** delete directories under `~/.cache`, e.g. `~/.cache/bundle`.

This is necessary to delete the bundle and bundle configuration files from the previous ~/.cache` location.

Corresponding PRs are https://github.com/typesafehub/conductr-cli/pull/298 and https://github.com/typesafehub/conductr-cli/pull/299.


markusjura:
Manually tested on macOS.

huntc (approved):

PR #326, opened 2017-02-24
Implements proxying feature that runs haproxy

longshorej:
Implements proxying feature that runs haproxy. This is part of the default set of features that can be disabled with `--no-default-features`. Features now have a `stop` method that will be called when `sandbox` stop is run, and proxying takes advantage of that.

### DIFF
--- a/conductr_cli/constants.py
+++ b/conductr_cli/constants.py
@@ -34,3 +34,7 @@ DEFAULT_SANDBOX_PROXY_CONTAINER_NAME = os.getenv('CONDUCTR_SANDBOX_PROXY_CONTAIN
 DEFAULT_ERROR_LOG_FILE = os.path.abspath(os.getenv('CONDUCTR_CLI_ERROR_LOG',
                                                    '{}/errors.log'.format(DEFAULT_CLI_SETTINGS_DIR)))
 DEFAULT_WAIT_TIMEOUT = 60  # seconds
+
+FEATURE_PROVIDE_PROXYING = 'proxying'
+
+FEATURE_PROVIDE_LOGGING = 'logging'
--- a/conductr_cli/sandbox_features.py
+++ b/conductr_cli/sandbox_features.py
@@ -15,16 +15,93 @@ Feature attributes:
 """
 
 import conductr_cli
-from conductr_cli import docker
+from conductr_cli import docker, sandbox_proxy
+from conductr_cli.constants import FEATURE_PROVIDE_LOGGING, FEATURE_PROVIDE_PROXYING
 from conductr_cli.sandbox_common import major_version
 from conductr_cli.screen_utils import h1
 import logging
+
+
+class FeatureStartResult:
+    def __init__(self, started, bundle_results):
+        self.started = started
+        self.bundle_results = bundle_results
 
 
 class BundleStartResult:
     def __init__(self, name, port):
         self.name = name
         self.port = port
+
+
+class ProxyingFeature:
+    """Proxying feature.
+
+    On start, the proxying features will be run.
+    """
+
+    name = 'proxying'
+    ports = []
+    dependencies = []
+    provides = [FEATURE_PROVIDE_PROXYING]
+
+    def __init__(self, version_args, image_version, offline_mode):
+        self.version_args = version_args
+        self.image_version = image_version
+        self.offline_mode = offline_mode
+        self.host = None
+        self.proxy_bind_addr = None
+        self.bundle_http_port = None
+        self.proxy_ports = None
+
+    def conductr_post_start(self, args, run_result):
+        self.host = run_result.host
+        self.proxy_bind_addr = run_result.core_addrs[0]
+        self.bundle_http_port = args.bundle_http_port
+        self.proxy_ports = sorted(args.ports)
+
+    @staticmethod
+    def conductr_args():
+        return []
+
+    def conductr_feature_envs(self):
+        return []
+
+    def conductr_roles(self):
+        if major_version(self.image_version) == 1:
+            return []
+        else:
+            return ['haproxy']
+
+    def start(self):
+        if major_version(self.image_version) == 1 or None in [self.host, self.proxy_bind_addr, self.bundle_http_port,
+                                                              self.proxy_ports]:
+            return FeatureStartResult(False, [])
+        else:
+            log = logging.getLogger(__name__)
+            log.info(h1('Starting proxying feature'))
+
+            started = sandbox_proxy.start_proxy(proxy_bind_addr=self.proxy_bind_addr,
+                                                bundle_http_port=self.bundle_http_port,
+                                                proxy_ports=self.proxy_ports,
+                                                all_feature_ports=all_feature_ports())
+
+            if started:
+                log.info('HAProxy has been started')
+                log.info('Your Bundles are by default accessible on:')
+                log.info('  {}:{}'.format(self.host, self.bundle_http_port))
+            else:
+                log.info('HAProxy has not been started')
+                log.info('To enable proxying ensure Docker is running and restart the sandbox')
+
+            return FeatureStartResult(started, [])
+
+    @staticmethod
+    def stop():
+        if sandbox_proxy.get_running_haproxy():
+            return sandbox_proxy.stop_proxy()
+        else:
+            return True
 
 
 class VisualizationFeature:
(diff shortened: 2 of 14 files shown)


conductr_cli/sandbox_run.py
```diff
@@ -42,15 +42,14 @@ def run(args):
...
-            print_proxy_output = sandbox_proxy.start_proxy(proxy_bind_addr=run_result.core_addrs[0],
-                                                           bundle_http_port=args.bundle_http_port,
-                                                           proxy_ports=sorted(args.ports))
-
         for feature in features:
+            feature.configure(args, run_result)
             result = feature.start()
             feature_results += result
 
+            if isinstance(feature, sandbox_features.ProxyingFeature):
```
longshorej: This I'm a bit iffy about, because Proxying is a "special" feature.

conductr_cli/constants.py
```diff
@@ -2,6 +2,8 @@
 
 CONDUCTR_SCHEME = 'CONDUCTR_SCHEME'
 
+CONDUCTR_BUNDLE_PROXY_PORT = 9000  # The default proxy port for bundles.
```
fsat: I think we would be able to make do without this constant. Refer to the comments below.

conductr_cli/sandbox_features.py
```diff
@@ -27,6 +28,73 @@ def __init__(self, name, port):
...
+            log.info(h1('Starting proxying feature'))
+
+            started = sandbox_proxy.start_proxy(proxy_bind_addr=self.proxy_bind_addr,
+                                                bundle_http_port=self.bundle_http_port,
+                                                proxy_ports=self.proxy_ports)
+
+            if started:
+                log.info('HAProxy has been started')
+                log.info('Your Bundles are by default accessible on:')
+                log.info('  {}:{}'.format(self.host, CONDUCTR_BUNDLE_PROXY_PORT))
```
fsat: I think `CONDUCTR_BUNDLE_PROXY_PORT` should be replaced in lieu of `self.bundle_http_port` as it's possible for the user to set bundle port apart from default `9000` and this value is kept in `self.bundle_http_port`.

conductr_cli/sandbox_run.py
huntc: Yeah, if feels wrong to do a special test. I thought that features had some facility to contribute to the output. Perhaps look at the monitoring feature for inspiration.

fsat: I'd suggest introduce a property `self.print_start_output = True` for the `ProxyingFeature`, and `self.print_start_output = False` for everything else. This way you can just check if `feature.print_start_output` is `True`, and print the output accordingly. Wdyt?

conductr_cli/test/test_sandbox_features.py
```diff
@@ -1,7 +1,7 @@
 from unittest import TestCase
 from unittest.mock import call, patch, MagicMock
 from conductr_cli.sandbox_features import VisualizationFeature, LiteLoggingFeature,\
-    LoggingFeature, MonitoringFeature, \
+    LoggingFeature, MonitoringFeature, ProxyingFeature, \
```
fsat: I'd suggest more coverage for `ProxyingFeature` in these areas:

- Output printout of during startup including when docker is enabled vs docker not enabled.
- Asserting the input argument and the call to `sandbox_proxy` during feature start depending of the input argument passed into `ProxyingFeature`'s `configure` method.
- Asserting the input argument and the call to `sandbox_proxy` during feature stop.

conductr_cli/test/test_sandbox_run.py
```diff
@@ -40,6 +40,26 @@ def default_general_args(self, container_name):
...
+    class DummyFeature2:
+        def __init__(self, version_args, image_version, offline_mode):
+            return
+
+        def configure(self, args, run_result):
+            return
+
+        def start(self):
+            return []
+
```
fsat: I can't seem to find the usage of `DummyFeature1` and `DummyFeature2`.

If they are meant to be used to assert that `configure` and `start` has been called, perhaps can be replaced with `MagicMock`, i.e.:

```
def create_dummy_features():
  feature = MagicMock()
  feature.configure = MagicMock()
  feature.start = MagicMock()
  return feature

feature_1 = self.create_dummy_features()
feature_2 = self.create_dummy_features()
features = [feature_1, feature_2] # use this as the return_value of mock_collect_features

feature_1.configure.assert_called_once_with(...)
feature_1.start.assert_called_once_with(...)
```

If they are indeed not used, can we remove them please?

fsat (changes requested):
Nice work!

I have some feedback about Proxying being a special feature, as well as some question about the new constant and test coverage for the new proxying feature.

conductr_cli/sandbox_features.py
```diff
@@ -27,6 +28,73 @@ def __init__(self, name, port):
...
+    def __init__(self, version_args, image_version, offline_mode):
+        self.version_args = version_args
+        self.image_version = image_version
+        self.offline_mode = offline_mode
+        self.host = None
+        self.proxy_bind_addr = None
+        self.bundle_http_port = None
+        self.proxy_ports = None
+
+    def configure(self, args, run_result):
```
markusjura: The name `configure` is quite general and does not indicate when it is called. What about `postConductrStart`?

markusjura (dismissed):
Looking good. Just one additional comment in regards to method naming.

conductr_cli/sandbox_run.py
longshorej: I think your suggestion to add a property to the feature is right but I'm thinking we can generalize it a bit.

As an idea, maybe features also have a `provides` property (e.g. the `proxy` and `log` class) such that only a single feature can be enabled at a time that has that `provides` entry. Then, this check simply becomes, were any features with `provides` `proxy` started? If so, we alter the service URL displayed. In the future, this would allow multiple feature implementations of the proxying concept and `sandbox_run_jvm.py` wouldn't care which was actually enabled.

Similarly, this would generalize the eslite/es logic. We could throw an error if two features were specified that both provided logging.

Just for reference, this boolean (`print_proxy_output`) is used as the following:

```
               for feature_result in feature_results:
                    if is_proxy_started:
                        uri = '{}:{}'.format(run_result.host, feature_result.port)
                    else:
                        uri = '{}:{}/services/{}'.format(run_result.host, DEFAULT_SERVICE_LOCATOR_PORT,
```

conductr_cli/constants.py
longshorej: Done.

conductr_cli/sandbox_features.py
longshorej: Done - per python conventions and other methods beginning with `conductr_`, went with `conductr_post_start`

conductr_cli/sandbox_features.py
longshorej: Done

conductr_cli/sandbox_run.py
longshorej: I went ahead with my idea here. Features now have a `provides` mechanism to provide any arbitrary string. No two enabled features can provide that same value -- this allows us to ensure only one logging feature is enabled, or in the future only one proxying feature. Let me know what you guys think of this!

conductr_cli/test/test_sandbox_features.py
longshorej: Good point - coverage was added for all of these cases except the first point. I moved the call to `sandbox_proxy.start` and there are tests on that method already to check the docker logic, so I think we're okay from a coverage standpoint.

conductr_cli/test/test_sandbox_run.py
longshorej: Done -- sorry, leftover from earlier iterations that I forgot to remove.

longshorej:
Okay, thanks @markusjura and @fsat for the feedback. I've incorporated it in the latest commit. To deal with the special case of the proxy being started affecting the URI calculation in `sandbox_run_jvm.log_run_attempt`, I introduced another concept for a `Feature` called `provides`. This is a list of arbitrary strings. `sandbox_run_jvm.log_run_attempt` then checks if any features provide `proxying` to calculate which URL to show.

This would allow us to swap out several different proxy features in the future, and allows us to introduce mutual exclusion for `logging` vs `lite-logging`.

I think it turned out nicely but let me know what you guys think.

fsat:
@longshorej - Thanks for incorporating my feedback. I've reviewed the PR, and I like the changes so far!

I do have a question about the `ProxyingFeature` test which I think it'll be faster if we can talk quickly after standup tomorrow.

fsat (approved):

PR #329, opened 2017-02-27
Calculate loopback interface name correctly

markusjura:
The loopback interface name always started with `lo0`, no matter if an alias with `lo0` already exists. This resulted into false output when someone first started the sandbox with `-n 1` and afterwards with a higher number, e.g. `-n3`.

Now the loopback interface name on linux is calculated based on the corresponding IP address. If the IP address is `192.168.0.2` then the loopback name is `lo1`.

This solution works both for IP4 and IP6.

### DIFF
--- a/conductr_cli/host.py
+++ b/conductr_cli/host.py
@@ -98,8 +98,8 @@ def addr_alias_setup_instructions(addrs, ip_version):
     subnet_mask = masks[ip_version]
 
     if is_linux():
-        commands = ['sudo ifconfig {}:{} {} netmask {} up'.format(if_name, idx, addr.exploded, subnet_mask)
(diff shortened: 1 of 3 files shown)


longshorej (approved):
I wonder if there's a way to actually test whether the loopback device is available? Another process could be using it and we'd be generating a command that'll fail.

markusjura:
@longshorej Well, we try to bind with a socket connection to certain ports and therefore check if the loopback device for this ports are available or not.

This would be certainly a separate issue / PR.

huntc:
I don't understand the motivation for this PR @markusjura. I would think that the interface should always be lo0. Was there a problem reported?

PR #332, opened 2017-03-01
WIP - DO NOT MERGE - idea for how shazar could calculate digest of stdin

longshorej:
This is a working example of how we could have `shazar` calculate digests of input streams. It's pretty simple - pipe stdin to stdout and calculate a running digest. Write this digest (32 bytes) to stdout when reading EOF of stdin.

Additional ideas to consider:

Perhaps `shazar verify` can verify if a file is valid: If it's a ZIP, check if its digest matches the one specified in the filename. If it's not a ZIP, calculate the digest excluding the last 32 bytes and see if the digest matches those bytes. Perhaps it has a `-q` flag where the only status indication is via exit code.

Example:

*Old Usage remains the same:*

```
$ shazar my-app
Created digested ZIP archive at ./my-app-8739c76e681f900923b900c9df0ef75cf421d39cabb54650c4b9ad19b6a76d85.zip
-> 0
```

*New usage activated when omitting a filename:*

```
$ echo 'hellofdsafdafs-world' | shazar
hellofdsafdafs-world[snippet some binary content]
-> 0
```

*New usage with `docker save` (note this would ultimately have `--format oci` but that's a WIP for them)*

```
$ docker save alpine:latest | shazar > my-shazared-oci-image.tar
-> 0

$ ls -lah my-shazared-oci-image.tar 
-rw-r--r-- 1 longshorej longshorej 4.1M Mar  1 17:37 my-shazared-oci-image.tar
-> 0
```

*Grabbing digest in shell*
```
$ tail -c 32 my-shazared-oci-image.tar | xxd -l 32 -c 32 -p
8e7d24db3adc0a3181419da17710d1790708b5e898157f2ffd288d9f9a2734fe
-> 0
```

### DIFF
--- a/conductr_cli/shazar_main.py
+++ b/conductr_cli/shazar_main.py
@@ -6,8 +6,12 @@ import hashlib
 import logging
 import os
 import shutil
+import sys
 import tempfile
 import zipfile
+
+
+READ_SIZE_BYTES = 32768
 
 
 def run(argv=None):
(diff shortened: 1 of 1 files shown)


huntc (approved):
I like it

longshorej:
Closing - superseded by the actual spec. Real implementation to come in separate PR.

Issue #340, opened 2017-03-09
conduct info does not work with released ConductR versions


markusjura:
The `conduct info` command is now retrieving the license from ConductR by using the `GET /license` endpoint. This endpoint is not available in the latest ConductR version 2.0.2. As a result, the following error is thrown:

```
$ conduct info
Error: 503 Service Unavailable
Error: Cannot presently locate the visualizer service - has it been started?
```

We need to change the conduct info code so that it supports all ConductR versions. So we should not do the license check for certain ConductR versions. This should be also considered in other areas of the conductr-cli where we call the new license endpoints.

markusjura:
![](https://github.trello.services/images/mini-trello-icon.png) [conductr-cli: conduct info should work for all ConductR versions](https://trello.com/c/d1NGjzyt/575-conductr-cli-conduct-info-should-work-for-all-conductr-versions)

huntc:
Yes, see https://github.com/typesafehub/sbt-conductr/pull/226#issuecomment-285325973

Tolerating 503 errors within conduct info should be sufficient.

PR #341, opened 2017-03-09
Catching license 503 errors

markusjura:
The license calls `GET /license` and `POST /license` catching now the HTTP status code 503 and then returning a `LICENSE_ENDPOINT_NOT_SUPPORTED` flag. Calling methods can then use this response to perform alternate logic.

The `conduct info` and `conduct load-license` commands have been modified accordingly.

**conduct info**
If a 503 is of `GET /license` is returned then no license info is printed. This is different to the return type None which means that no license has been uploaded to ConductR which will print the no license information

**conduct load-license**
Before posting a license we now perform a `GET /license` call to check if a 503 is returned. If so the message `conduct load-license is only supported by ConductR 2.0.3+` is printed to the console. Note that this message can be suppressed if `args.quite` is et to True. This might be helpful if the function is called from another Python function that don’t want to print any license output for ConductR versions 2.0.2-

Fixes https://github.com/typesafehub/conductr-cli/issues/340

### DIFF
--- a/conductr_cli/license.py
+++ b/conductr_cli/license.py
@@ -4,7 +4,6 @@ import arrow
 import datetime
 import json
 
-
 EXPIRY_DATE_DISPLAY_FORMAT = '%a %d %b %Y %H:%M%p'
 
 UNLICENSED_DISPLAY_TEXT = 'UNLICENSED - please use "conduct load-license" to use more than one agent. ' \
@@ -34,8 +33,11 @@ def post_license(args, license_file):
                                     data=open(license_file, 'rb'),
                                     auth=args.conductr_auth,
                                     verify=args.server_verification_file)
-    validation.raise_for_status_inc_3xx(response)
-    return True
+    if response.status_code == 503:
+        return False
+    else:
+        validation.raise_for_status_inc_3xx(response)
+        return True
 
 
 def format_license(license):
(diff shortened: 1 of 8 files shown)


conductr_cli/license.py
```diff
@@ -34,8 +35,11 @@ def post_license(args, license_file):
                                     data=open(license_file, 'rb'),
                                     auth=args.conductr_auth,
                                     verify=args.server_verification_file)
-    validation.raise_for_status_inc_3xx(response)
-    return True
+    if response.status_code == 503:
+        return LICENSE_ENDPOINT_NOT_SUPPORTED
```
longshorej: Does it make sense to just return `False` instead?

conductr_cli/license.py
```diff
@@ -78,10 +82,16 @@ def format_multiple(key, title_text):
 def get_license(args):
```
longshorej, line 80: I think usage would be clearer if we returned a 2 value tuple - (forgive the Scala type declaration :p)
   `(Boolean, Option[Json])` - and then we wouldn't need a special constant either.

So not supported returns: `False, None`
Supported, but no data: `True, None`
Supported with data: `True, Json`

longshorej (approved):
Approving but with some comments on return types that I think could be simplified.

conductr_cli/license.py
markusjura: Addressed you PR feedback for both get_license and post_license.

PR #358, opened 2017-03-20
Hostname lookup validation

markusjura:
Adding additional check how long the hostname lookup takes. This validation is necessary to check if ConductR would fail at startup given this issue with Java 8 on macOS: https://github.com/akka/akka/issues/22160

The validation is calling a simple jar files that outputs the lookup time of `java.net.InetAddress.getLocalHost(InetAddress.java:1500)`. If the lookup takes longer than 3 seconds then the validation fails and the following validation error is thrown:

```
$ sandbox run 2.0.2
Error: Hostname lookup took 5006 milliseconds which will result in a ConductR startup failure
Error: This is known Java 8 issue on macOS: http://stackoverflow.com/questions/39636792/jvm-takes-a-long-time-to-resolve-ip-address-for-localhost
Error: To speed up the hostname lookup add your macOS hostname to /etc/hosts
Error: Resolves your hostname on the terminal with: hostname
Error: Sample /etc/hosts file:
Error: 127.0.0.1   localhost mbpro.local
Error: ::1         localhost mbpro.local
```

If the hostname is set in the `/etc/hosts` file then the lookup takes about 30 ms.

The advantage of calling the jar file to check the hostname lookup time is that once the issue has been fixed by either a new Java version or a new macOS version then this validation will pass. Also, without actually checking the lookup time it is hard to validate it because this problem can occur on different macOS and Java versions.

The check is only performed in sandbox JVM mode.


setup.py
```diff
@@ -64,6 +64,10 @@ def run_tests(self):
     version=__version__,
     packages=find_packages(),
 
+    data_files=[
```
markusjura, line 67: @huntc @fsat @longshorej Does adding the binary file with `data_files` work for the native and pip3 distribution?

huntc: Easy to check... just invoke pyinstaller :-)

huntc (commented):
It still concerns me that we have to work around this OS X issue. Surely there must have been some progress in order to fix it...

setup.py
markusjura: It is working for the native install:

```
/Users/mj/workspace/conductr-cli/dist/sandbox run 2.0.2
Error: Hostname lookup took 5008 milliseconds which will result in a ConductR startup failure
Error: This is known Java 8 issue on macOS: http://stackoverflow.com/questions/39636792/jvm-takes-a-long-time-to-resolve-ip-address-for-localhost
Error: To speed up the hostname lookup add your macOS hostname to /etc/hosts
Error: Resolves your hostname on the terminal with: hostname
Error: Sample /etc/hosts file:
Error: 127.0.0.1   localhost mbpro.local
Error: ::1         localhost mbpro.local
```

huntc (approved):
Please document the source of the jar - perhaps in our readme, or a file closely associated with the jar.

bin/README.md
```diff
@@ -0,0 +1,9 @@
+# Hostname Lookup Tester
```
markusjura, line 1: Added this README to document the source of the `inetTester.jar`.

PR #359, opened 2017-03-21
Corrected dev version check

huntc:
0.1.0 bundles should enable proxying, this is a regression.


huntc:
Tested locally. All is now well for 0.1.0 and proxying.

fsat (approved):

PR #363, opened 2017-03-22
Revert "Hostname lookup validation"

markusjura:
Reverting hostname lookup validation of PR https://github.com/typesafehub/conductr-cli/pull/358 for now because this is working for the native distribution but not for the `pip3` distribution.

Will create another PR that will work for both distributions.


markusjura:
Tested the native and pip3 distribution based on this PR

PR #364, opened 2017-03-22
Fix tests failing with `python3 -m unittest`

longshorej:
It seems `tox` and `python3 -m unittest` are slightly different. `tox` will be okay with tests that mock `sys.stdout.buffer` but `python3 -m unittest` will fail saying it's a readonly attribute. This PR adjusts travis to run tests both ways and fixes those that were failing.

This PR also included manually testing travis by only updating `travis.yml`. When I did that, I was able to verify that the tests were indeed failing. Then, I pushed another commit to get them running correctly and squashed commits, so this is ready for merge upon approval.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ install:
   - pip install virtualenv
   - pip install tox
 script:
+  - pip3 install -e .
   - tox
+  - python3 -m unittest
 deploy:
   provider: pypi
   user: typesafehub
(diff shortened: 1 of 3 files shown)


longshorej:
Thanks to @markusjura for letting me know about `python3 -m unittest`!

.travis.yml
```diff
@@ -5,7 +5,9 @@ install:
   - pip install virtualenv
   - pip install tox
 script:
+  - pip3 install -e .
```
huntc, line 8: This shouldn't be necessary

huntc (approved):

.travis.yml
longshorej: It is - `tox` installs dependencies for itself but `python3` doesn't have access to those. Without this line, `python3 -m unittest` fails due to dependency issues.

PR #365, opened 2017-03-22
Add options to `bndl` to allow adjustment of docker image prior to conversion

longshorej:
This commit adds a couple useful options to `bndl` to allow you to transform the docker image before it is turned into a bundle -- overwriting `Cmd`, `Entrypoint`, and `Env` is possible.

Because of how these values are used, they accept an argument in JSON format. Otherwise, there are ambiguities with argparse. For example, flags meant to be part of `Cmd` cannot easily be expressed.

**Manual Test:**

```bash
$ docker save wstest:latest | bndl --no-shazar --docker-cmd '["/hello/world"]' --docker-env '["PATH=/"]' --docker-entrypoint '["/bin/bash"]' | tar x 
-> 0

$ cat wstest/oci-image/blobs/sha256/14cc8e5430e5f7e30bee8f024e008f374d307879a5503b7ccf2d4f3a8336a841  | jsl-format-json 
{
    "architecture": "amd64",
    "config": {
        "Cmd": [
            "/hello/world"
        ],
        "Entrypoint": [
            "/bin/bash"
        ],
        "Env": [
            "PATH=/"
        ]
    },
   ...
}

### DIFF
--- a/conductr_cli/bndl_create.py
+++ b/conductr_cli/bndl_create.py
@@ -48,15 +48,37 @@ def bndl_create(args):
 
             return 2
         elif args.format == 'docker':
-            if args.source is None:
-                with tarfile.open(fileobj=buff_in, mode='r|') as tar_in:
-                    name = docker_unpack(oci_image_dir, tar_in, is_dir=False, maybe_name=args.name, maybe_tag=args.tag)
-            elif os.path.isfile(args.source):
-                with tarfile.open(args.source, mode='r') as tar_in:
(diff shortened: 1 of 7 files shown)


huntc:
This could be a slippery slope and doesn't feel within the scope of bndl. I'd suggest researching other Docker tooling to achieve your goal.

huntc (changes requested):
Please see other comment

longshorej:
Closed due to scope creep.

PR #372, opened 2017-03-27
Hostname lookup validation pre-check

markusjura:
Adding additional check that the hostname lookup doesn’t take too long. This validation is necessary to check if ConductR would fail at startup given this issue with Java 8 on macOS: https://github.com/akka/akka/issues/22160

The validation is failing if the hostname is not part of the `/etc/hosts` file and if the operation system is macOS. In this case, we abort with this error:

```
$ sandbox run 2.0.5
Error: Hostname lookup on your machine will take more than 5 seconds which will result in a ConductR startup failure
Error: This is known Java 8 issue on macOS: http://stackoverflow.com/questions/39636792/jvm-takes-a-long-time-to-resolve-ip-address-for-localhost
Error: To speed up the hostname lookup add your macOS hostname to /etc/hosts
Error: Resolves your hostname on the terminal with: hostname
Error: Sample /etc/hosts file:
Error: 127.0.0.1   localhost mbpro.local
Error: ::1         localhost mbpro.local
```

We do not check for the Java version given that this error only occurs from Java 8 onwards and ConductR supports only Java 8 and onwards.

The check is only performed in sandbox JVM mode.

### DIFF
--- a/conductr_cli/sandbox_run_jvm.py
+++ b/conductr_cli/sandbox_run_jvm.py
@@ -4,7 +4,7 @@ from conductr_cli.constants import DEFAULT_SCHEME, DEFAULT_PORT, DEFAULT_BASE_PA
 from conductr_cli.exceptions import BindAddressNotFound, BintrayUnreachableError, InstanceCountError, \
     SandboxImageNotFoundError, SandboxImageNotAvailableOfflineError, SandboxUnsupportedOsArchError, \
     SandboxUnsupportedOsError, JavaCallError, JavaUnsupportedVendorError, JavaUnsupportedVersionError, \
-    JavaVersionParseError, LicenseValidationError
+    JavaVersionParseError, HostnameLookupError, LicenseValidationError
 from conductr_cli.resolvers import bintray_resolver
 from conductr_cli.resolvers.bintray_resolver import BINTRAY_LIGHTBEND_ORG, BINTRAY_CONDUCTR_REPO
 from conductr_cli.sandbox_common import flatten
@@ -73,7 +73,7 @@ def run(args, features):
     nr_of_core_instances, nr_of_agent_instances = instance_count(args.image_version, args.nr_of_instances)
 
     validate_jvm_support()
-
+    validate_hostname_lookup()
     validate_64bit_support()
 
     sandbox_stop.stop(args)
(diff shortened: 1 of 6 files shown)


conductr_cli/sandbox_run_jvm.py
```diff
@@ -235,6 +235,22 @@ def validate_jvm_support():
         raise JavaCallError('Failure calling `java -version`')
 
 
+def validate_hostname_lookup():
+    """
+    Validates if the hostname lookup by Java is slow given the issue https://github.com/akka/akka/issues/22160
+    Validation fails if the hostname is not part of the /etc/hosts file and the OS is macOS.
+    A HostnameLookupError is raised if the validation fails
+    """
+    if host.is_macos():
```
huntc, line 244: I do think we should check the Java version also as the CLI can operate on a machine where non ConductR dev occurs.

conductr_cli/validation.py
```diff
@@ -438,6 +438,31 @@ def handler(*args, **kwargs):
...
+def handle_hostname_lookup_error(func):
+
+    def handler(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except HostnameLookupError:
+            log = get_logger_for_func(func)
+            log.error('Hostname lookup on your machine will take more than 5 seconds '
+                      'which will result in a ConductR startup failure')
+            log.error('This is known Java 8 issue on macOS: http://stackoverflow.com/questions/39636792/'
```
huntc, line 450: "Mac OS X"

huntc (changes requested):

conductr_cli/validation.py
longshorej: Isn't it "macOS" now? http://www.apple.com/macos/sierra/

huntc: I'm starting to feel old now. You both win. :-)

markusjura: Yes, we also using `macOS` in every other place.

conductr_cli/sandbox_run_jvm.py
markusjura: We validate the Java version prior to this check. So if the Java version is not valid then we do not we will abort before we perform the hostname lookup validation.

huntc (approved):

PR #373, opened 2017-03-27
bndl expanded to parse OCI `ExposedPorts`

longshorej:
**Updated 2017-03-27 11:17 PM UTC**

This PR modifies `bndl` so that it creates `endpoint` entries in `bundle.conf` for OCI Image bundles that contain `ExposedPorts` declarations. It also creates a status component that runs `check` on the appropriate addresses. Lastly, it fixes an issue where layers that were symlinks weren't resolved properly.

A user can disable this behavior with the `--no-default-endpoints` flag but I think it's reasonable and indeed expected to be enabled by default.

### Manual Tests

```bash
$ docker save tutum/hello-world | bndl --no-shazar | tar x
-> 0

$ cat hello-world/bundle.conf 
name = "hello-world"
roles = []
compatibilityVersion = 0
diskSpace = 1073741824
memory = 402653184
nrOfCpus = 0.1
system = "hello-world"
version = 1
components {
  oci-image {
    description = ""
    file-system-type = "oci-image"
    start-command = [
      "ociImageTag"
      "latest"
    ]
    endpoints {
      oci-image-tcp-80 {
        bind-protocol = "tcp"
        bind-port = 80
        service-name = "oci-image-tcp-80"
      }
    }
  }
  oci-image-status {
    description = "Status check for oci-image component"
    file-system-type = "universal"
    start-command = [
      "check"
      "$OCI_IMAGE_TCP_80_HOST"
    ]
    endpoints {}
  }
}-> 0



docker save tutum/hello-world | bndl --no-default-endpoints --no-shazar | tar x
-> 0

cat hello-world/bundle.conf 
name = "hello-world"
roles = []
compatibilityVersion = 0
diskSpace = 1073741824
memory = 402653184
nrOfCpus = 0.1
system = "hello-world"
version = 1
components {
  oci-image {
    description = ""
    file-system-type = "oci-image"
    start-command = [
      "ociImageTag"
      "latest"
    ]
    endpoints {}
  }
}-> 0
```

### DIFF
--- a/conductr_cli/bndl_create.py
+++ b/conductr_cli/bndl_create.py
@@ -1,4 +1,4 @@
-from conductr_cli.bndl_oci import oci_image_bundle_conf, oci_image_unpack
+from conductr_cli.bndl_oci import oci_image_bundle_conf, oci_image_extract_config, oci_image_unpack
 from conductr_cli.bndl_docker import docker_unpack
 from conductr_cli.bndl_utils import detect_format_dir, detect_format_stream
 from conductr_cli.constants import BNDL_PEEK_SIZE, IO_CHUNK_SIZE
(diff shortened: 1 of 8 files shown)


huntc (changes requested):
Nice idea. We should discuss this further as I think 80 and 443 can be assumed as http/s by default, and we could use the bundle name to discriminate paths, again by default.

huntc (approved):
Looking great - ignore the http comment from before. As per conversation, if we generate endpoint code then we should also generate a check command. This keeps the "pure Docker" approach pretty sweet.

Thanks for the awesome work!

fsat (approved):

PR #382, opened 2017-03-30
Brew PR command

huntc:
Perform the brew PR command in order to make generating the brew distribution a bit easier.

### DIFF
--- a/package-native-zip.sh
+++ b/package-native-zip.sh
@@ -14,6 +14,11 @@ UNAME_VALUE=`uname`
 
 if [ "$UNAME_VALUE" = "Darwin" ]; then
     PACKAGE_OS_NAME="Mac_OS_X-x86_64"
+
+    if ! brew -v >/dev/null 2>&1; then
+        echo "ERROR: brew must be installed in order to package "
+        exit 1
+    fi
 elif [ "$UNAME_VALUE" = "Linux" ]; then
     PACKAGE_OS_NAME="Linux-amd64"
 else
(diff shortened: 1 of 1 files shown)


package-native-zip.sh
```diff
@@ -103,6 +103,23 @@ echo "------------------------------------------------"
 echo ""
 zip -j dist/$PACKAGE_NAME dist/conduct dist/sandbox dist/shazar
 
+if [ "$UNAME_VALUE" = "Darwin" ]; then
+    echo "------------------------------------------------"
+    echo "Commands to create the Homebrew package pull request for $PACKAGE_NAME"
+    echo "------------------------------------------------"
+    echo ""
+
+    echo "Checkout the repo at https://github.com/typesafehub/homebrew-conductr and then issue"
```
TimMoore: I don't think you need to have the repo cloned, but you _do_ need to have it "tapped" in your Homebrew installation, so I'd say: "Run `brew tap typesafehub/conductr` and then issue"...

TimMoore: Or just add it to the following brew commands.

TimMoore: Looks like you also need to have `hub` installed and configured. I already did, so it just worked for me, but it might be worth noting here. You can install it with Homebrew (`brew install hub`) and IIRC it prompts for credentials when you first use it.

https://github.com/Homebrew/brew/blob/e022239608ca535fa3b56305eff8dd091b3fc008/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L303-L304

TimMoore (approved):

PR #383, opened 2017-03-30
Incorporate PID output

huntc:
ConductR 2.1 will represent an OS process identifier (PID) against each bundle execution. This commit caters for both ConductR 2.1 and existing ConductR releases by outputting a PID during "conduct info <my-bundle>" if there is one. If there is no PID field then "Unknown" is output.

### DIFF
--- a/conductr_cli/conduct_info_inspect.py
+++ b/conductr_cli/conduct_info_inspect.py
@@ -134,6 +134,7 @@ def display_bundle_executions(bundle):
             {
                 'endpoint': endpoint_name,
                 'host': bundle_execution['host'],
+                'pid': bundle_execution['pid'] if 'pid' in bundle_execution else 'Unknown',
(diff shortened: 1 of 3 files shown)


typesafe-tools:
[conduct info <bundle> should display pid if it is available as a field.](https://trello.com/c/J6nX8uvx/611-conduct-info-bundle-should-display-pid-if-it-is-available-as-a-field)

conductr_cli/conduct_info_inspect.py
```diff
@@ -156,6 +158,7 @@ def display_bundle_executions(bundle):
             for row in rows:
                 log.screen('{endpoint: <{endpoint_width}}{padding}'
                            '{host: <{host_width}}{padding}'
+                           '{pid: <{pid_width}}{padding}'
```
markusjura: pid is a number and therefore should be right aligned, e.g. as `>{bind_port_width}`

markusjura (dismissed):

PR #386, opened 2017-04-03
Display uptime on conduct info detail page

markusjura:
The `/bundles` endpoints contain the `start_time` of a bundle execution. We now calculate the uptime of a bundle based on the start time and the current time and display it on the `conduct info my-bundle` detail page:

```
$ conduct info eslite
BUNDLE ATTRIBUTES
-----------------
Bundle Id              3349b6b
Bundle Name            eslite
Compatibility Version  1
System                 eslite
System Version         1
Tags
Nr of CPUs             0.1
Memory                 402653184
Disk Space             200000000
Roles                  elasticsearch
Bundle Digest          3349b6bfbc09e8d88123fee5b33b2717e07fb2a91de719c99e0403c9e4b28ea5
Error                  No

BUNDLE SCALE
------------
Nr of Reschedules  0
Scale              1

BUNDLE INSTALLATIONS
--------------------
Host    192.168.10.1
Bundle  /Users/mj/.conductr/images/tmp/conductr/192.168.10.1/bundles/3349b6bfbc09e8d88123fee5b33b2717e07fb2a91de719c99e0403c9e4b28ea5.zip

BUNDLE EXECUTIONS
-----------------
ENDPOINT     HOST            PID  STARTED        UPTIME  BIND_PORT  HOST_PORT
akka-remote  192.168.10.1  26793      Yes  2h, 38m, 48s      10822      10822
es           192.168.10.1  26793      Yes  2h, 38m, 48s      10007      10007

SERVICE NAMES
-------------
SERVICE NAME    STATUS
elastic-search  Running
```

Displaying how long a bundle is running (uptime) is more meaningful in this context as displaying the start time. Displaying both info makes the row unnecessary long.

If the `start_time` key in the JSON response does not exists than we display `Unknown`.

The uptime format is humanized. It displays the time as `10 d, 2h, 38m, 48s`. If the delta is less than a day, an hour, a minute, then these info's are not displayed, e.g.

```
BUNDLE EXECUTIONS
-----------------
ENDPOINT     HOST            PID  STARTED                 UPTIME  BIND_PORT  HOST_PORT
akka-remote  192.168.10.1  26793      Yes  10 days, 2h, 38m, 48s      10822      10822
```

```
BUNDLE EXECUTIONS
-----------------
ENDPOINT     HOST            PID  STARTED    UPTIME  BIND_PORT  HOST_PORT
akka-remote  192.168.10.1  26793      Yes  38m, 48s      10822      10822
```

```
BUNDLE EXECUTIONS
-----------------
ENDPOINT     HOST            PID  STARTED  UPTIME  BIND_PORT  HOST_PORT
akka-remote  192.168.10.1  26793      Yes     48s      10822      10822
```

### DIFF
--- a/conductr_cli/conduct_info_inspect.py
+++ b/conductr_cli/conduct_info_inspect.py
@@ -1,6 +1,7 @@
 from conductr_cli import screen_utils, conduct_acls, conduct_service_names
 from conductr_cli.conduct_info_common import DISPLAY_PADDING, display_bundle_id
 from urllib.parse import urlparse
+import arrow
 import json
 import logging
 
@@ -137,6 +138,7 @@ def display_bundle_executions(bundle):
                 'host': bundle_execution['host'],
                 'pid': bundle_execution['pid'] if 'pid' in bundle_execution else 'Unknown',
                 'is_started': 'Yes' if bundle_execution['isStarted'] else 'No',
+                'uptime': get_uptime(bundle_execution['startTime']) if 'startTime' in bundle_execution else 'Unknown',
                 'bind_port': endpoint_details['bindPort'],
                 'host_port': endpoint_details['hostPort']
             }
@@ -152,6 +154,7 @@ def display_bundle_executions(bundle):
                 'host': 'HOST',
                 'pid': 'PID',
                 'is_started': 'STARTED',
+                'uptime': 'UPTIME',
                 'bind_port': 'BIND_PORT',
                 'host_port': 'HOST_PORT',
             })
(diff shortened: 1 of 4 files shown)


conductr_cli/conduct_info_inspect.py
```diff
@@ -206,6 +210,32 @@ def display_title_table(title):
...
+            if value:
+                seconds -= value * count
+                # if value > 0:
+                if value == 1:
+                    name = name.rstrip('s')
+                result.append("{}{}".format(value, name))
+        return ', '.join(result)
+
+    start_time_local = arrow.get(start_time).to('local')
+    current_time_local = arrow.now().to('local')
```
longshorej: I think we need to do the seconds calculation with UTC. Local timezones don't necessary have 86400 seconds in them due to DST but POSIX time always does. Also there's the consideration of clocks being set incorrectly, we probably want to make sure we don't show a negative uptime value.

Perhaps `seconds = max(0, arrow.now().timestamp - arrow.get(start_time).timestamp)`

longshorej (changes requested):
Looks good, just one comment on `get_uptime` calculation.

longshorej (approved):

PR #393, opened 2017-04-04
Add bndl to package-native script

markusjura:


longshorej (approved):

huntc (approved):

huntc:
Changing from WIP as @longshorej confirmed that Windows builds are fine (which I believe is why this was left as a WIP).

markusjura:
The assumption is correct. The PR was only marked as WIP due to the native scripts were not working on Windows (for me).

PR #396, opened 2017-04-05
Adjusts the sandbox to pull down the OCI in Docker image when starting

longshorej:
**Updated 04/06/2017 4:14pm UTC**

This PR makes the sandbox pull down the relevant image for running OCI in Docker when it starts up. 

The code to do this has been implemented in a feature and that feature is enabled by default in non-linux environments. It can thus be enabled on linux via `sandbox run 2.0.5 --feature oci-in-docker`.

It determines the image to pull down by reading it from the agent's `application.conf`. If the value is not present, we assume we're in a pre-OCI version of ConductR and don't do anything.

### DIFF
--- a/conductr_cli/sandbox_features.py
+++ b/conductr_cli/sandbox_features.py
@@ -21,11 +21,13 @@ Feature attributes:
 """
 
 import conductr_cli
-from conductr_cli import docker, sandbox_proxy
+from conductr_cli import docker, host, sandbox_proxy, terminal
 from conductr_cli.sandbox_version import is_conductr_supportive_of_features, is_cinnamon_grafana_docker_based
 from conductr_cli.constants import FEATURE_PROVIDE_LOGGING, FEATURE_PROVIDE_PROXYING
 from conductr_cli.screen_utils import h1
+from zipfile import ZipFile
 import logging
+import os
 
 
 class FeatureStartResult:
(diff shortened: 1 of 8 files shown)


longshorej:
Also relevant is this repo I created today for building this image. It's currently very simple but I expect we'll add glibc at some point soon: https://github.com/typesafehub/conductr-oci-in-docker

huntc:
Can we not store the dockerfile that is required in the lib folder of an agent image and then use that for building?

longshorej:
I think that strategy would work but it may not help our goal here. We want to reduce the amount of time it takes to launch the sandbox to a very minimal amount of time. I think the best strategy for that is for us to build the images and host them on bintray so that the users of the sandbox simply have to pull them down. They don't have to wait for a `docker build` then. wdyt?

huntc:
> https://github.com/typesafehub/conductr-cli/pull/396#issuecomment-291994068

Sounds great.

conductr_cli/sandbox_features.py
```diff
@@ -468,16 +548,23 @@ def collect_features(feature_args, no_default_features, image_version, offline_m
...
 
         def add_logging_lite(features):
-            names = [feature.name for feature in features]
             if LoggingFeature.name not in names:
                 features.insert(0, feature_lookup[LiteLoggingFeature.name]([], image_version, offline_mode))
 
+        def add_oci_in_docker(features):
+            if OciInDockerFeature.name not in names and not host.is_linux():
+                features.insert(0, feature_lookup[OciInDockerFeature.name]([], image_version, offline_mode))
+
```
huntc, line 587: I wonder if there is a use-case where we don't want to run OCI images... perhaps not as we want almost everything to run via OCI.

huntc (approved):

conductr_cli/sandbox_features.py
longshorej: Hrm, not sure if there is, but since it's a feature this code can now be disabled with `--no-default-features` so that might satisfy the use-case.

markusjura (approved):

huntc (approved):

PR #397, opened 2017-04-06
Handle stopping state

huntc:
In ConductR 2.1, a bundle may now be stopping and thus set inActive (isStarted) to false... In this case, when a "conduct stop" command is issue, we want to wait for bundle execution entry to actually disappear before being out of the scale count i.e. we are only waiting on isActive to become True and thus to be part of the scale count when scaling up.

### DIFF
--- a/conductr_cli/conduct_run.py
+++ b/conductr_cli/conduct_run.py
@@ -34,7 +34,7 @@ def run(args):
     log.info('Bundle run request sent.')
 
     if not args.no_wait:
-        bundle_scale.wait_for_scale(response_json['bundleId'], args.scale, args)
+        bundle_scale.wait_for_scale(response_json['bundleId'], args.scale, wait_for_is_active=True, args=args)
 
     if not args.disable_instructions:
         log.info('Stop bundle with:         {} stop{} {}'.format(args.command, args.cli_parameters, bundle_id))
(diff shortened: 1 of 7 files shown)


conductr_cli/conduct_run.py
```diff
@@ -34,7 +34,7 @@ def run(args):
     log.info('Bundle run request sent.')
 
     if not args.no_wait:
-        bundle_scale.wait_for_scale(response_json['bundleId'], args.scale, args)
+        bundle_scale.wait_for_scale(response_json['bundleId'], args.scale, True, args)
```
markusjura: Use `wait_for_is_active=True`

conductr_cli/conduct_stop.py
```diff
@@ -27,7 +27,7 @@ def stop(args):
     log.info('Bundle stop request sent.')
 
     if not args.no_wait:
-        bundle_scale.wait_for_scale(response_json['bundleId'], 0, args)
+        bundle_scale.wait_for_scale(response_json['bundleId'], 0, False, args)
```
markusjura: Use `wait_for_is_active=False`

conductr_cli/conduct_run.py
markusjura: Note that you also need to change the tests accordingly.

markusjura (approved):
Approved, but small changes necessary.

typesafe-tools:
[Fix a problem in the CLI where there's a race between stopping and unload](https://trello.com/c/5f9vsjqY/640-fix-a-problem-in-the-cli-where-there-s-a-race-between-stopping-and-unload)

PR #401, opened 2017-04-07
Adds `brew install python3` for OSX users.

### LINKED ISSUE #400: Docs: in OSX python3 is required by cassandra bundle
If using the `native` approach to installing conductr-cli on OSX, it's possible the host machine doesn't have python3 installed.

Cassandra bundle won't start if `python3` can't be found.

ignasi35:
Fixes #400.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,6 +19,12 @@ If you use macOS and `brew` then you can quickly:
 .. code:: bash
     
     brew install typesafehub/conductr/conductr-cli
+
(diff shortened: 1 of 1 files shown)


huntc (commented):
Python3 shouldn't be required by our bundles... we should fix them.

huntc:
Closing here. @markusjura, can we please investigate why our Cassandra bundle depends on python3? Obviously something I assumed a while back. Note that this problem can be alleviated with our move toward OCI also... thanks.

ignasi35:
Sorry I didn't catch the traces. I was using `sandbox logs -f` and we I noticed I should have copy/pasted the logs had overflowed to buffer.

PR #403, opened 2017-04-08
Fix version check

huntc:
Version checking wasn't supporting our semantic versions i.e. 2.1.0-alpha.1.

Also fixed a problem when testing features on OS X - the tests didn't expect the OCI feature.


huntc:
Self merging - need to get this release out and the previous version check logic prevented us from doing that.

Issue #416, opened 2017-04-17
Hostname lookup validation (for slow startup on macOS) fails in certain situations


longshorej:
1. Comments are only respected as the first character of the line -- for instance, this would fail and erroneously allow the validation to pass: `127.0.0.1 localhost # My-Mac.local`

2. All lines that mentions `localhost` are required to have the hostname in it or the validation will fail. I think we want to invert the logic - if any lines with `localhost` in it also have the hostname, we should pass. We'd need someone with a mac to test this assertion out, however.
Below is @dietrichs failing /etc/hosts file that succeeded when adding the hostname to the last line.
```
127.0.0.1        localhost Dietrichs-MBP.local monitor.prod.opsclarity.com
::1             localhost Dietrichs-MBP.local
255.255.255.255    broadcasthost
fe80::1%lo0        localhost
```

PR #420, opened 2017-04-18
Hostname required in only one of localhost line

markusjura:
The hostname lookup validation now expects the hostname in only one of the localhost lines in `/etc/hosts`. Also, we now consider comments within a line in `/etc/hosts`, e.g.

```
127.0.0.1	localhost # mj-mbpro.local
```

Fixes https://github.com/typesafehub/conductr-cli/issues/416.

Tested all scenarios on macOS and added respective unit tests.

### DIFF
--- a/conductr_cli/sandbox_run_jvm.py
+++ b/conductr_cli/sandbox_run_jvm.py
@@ -255,17 +255,17 @@ def validate_jvm_support():
 def validate_hostname_lookup():
     """
     Validates if the hostname lookup by Java is slow given the issue https://github.com/akka/akka/issues/22160
-    Validation fails if the hostname is not part of the /etc/hosts file and the OS is macOS.
+    Validation fails if the hostname is not part of at least one localhost entry in the /etc/hosts file
+    and the OS is macOS.
     A HostnameLookupError is raised if the validation fails
     """
     if host.is_macos():
         with open('/etc/hosts', 'r') as file:
             hostname = host.hostname()
-            lines = [line for line in file.readlines() if not line.strip().startswith('#')]
+            lines = [line.split('#')[0] for line in file.readlines() if not line.strip().startswith('#')]
             localhost_lines = [line for line in lines if 'localhost' in line]
-            for line in localhost_lines:
-                if hostname not in line:
-                    raise HostnameLookupError()
+            if not any(hostname in words for words in (line.split() for line in localhost_lines)):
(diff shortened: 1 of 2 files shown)


conductr_cli/sandbox_run_jvm.py
```diff
@@ -255,17 +255,17 @@ def validate_jvm_support():
...
     if host.is_macos():
         with open('/etc/hosts', 'r') as file:
             hostname = host.hostname()
-            lines = [line for line in file.readlines() if not line.strip().startswith('#')]
+            lines = [line.split('#')[0] for line in file.readlines() if not line.strip().startswith('#')]
             localhost_lines = [line for line in lines if 'localhost' in line]
-            for line in localhost_lines:
-                if hostname not in line:
-                    raise HostnameLookupError()
+            if not any(hostname in line for line in localhost_lines):
```
longshorej: I think we need to split by white-space as well. Consider the hostname being equal to `lightbend1` but there's an entry for `lightbend11` - I think this will be a false positive match. Granted it's unlikely to happen in the sandbox but still possible.

Perhaps something like:
```if not any any(hostname in words for words in (line.split() for line in localhost_lines)):```

File: conductr_cli/sandbox_run_jvm.py [hunk 2]
```diff
@@ -255,17 +255,17 @@ def validate_jvm_support():
...
-    Validation fails if the hostname is not part of the /etc/hosts file and the OS is macOS.
+    Validation fails if the hostname is not part of at least one localhost entry in the /etc/hosts file 
+    and the OS is macOS.
     A HostnameLookupError is raised if the validation fails
     """
     if host.is_macos():
         with open('/etc/hosts', 'r') as file:
             hostname = host.hostname()
-            lines = [line for line in file.readlines() if not line.strip().startswith('#')]
+            lines = [line.split('#')[0] for line in file.readlines() if not line.strip().startswith('#')]
```
longshorej, line 265: 👍

longshorej (changes requested):
Looks good overall. I realized there was one more case that I think we should address.

conductr_cli/sandbox_run_jvm.py
markusjura: Good catch. Changed it as you've suggested and added a unit test for this case.

longshorej (approved):

PR #438, opened 2017-04-30
Upgrading dcos modules to 0.4.16

markusjura:
Upgrading the embedded DC/OS files to dcos-cli version 0.4.16. This version allows connections to DC/OS clusters that have authentication disabled.

Due to this upgrade the following changes in the conductr-cli where necessary:
- Updated requirements in setup.py, e.g. jsonschema > 2.5
- Exception handling of DCOSHttpException. Now we only log the output of `__str__` instead of the `err.response`, `err. err.response.status` and `err. err.response .text`. This is because the `DCOSAuthenticationException` does not contain the attribute `reason`.
- In DC/OS mode, the ConductR url does not contain the port. Without this change, the `dcos conduct info` returns a `401 Not Authorized` because in the DC/OS http library, a url match is performed between the passed in URL and the URL configured with `dcos config`. The parsed `dcos config` URL does not contain a port information: https://github.com/dcos/dcos-cli/blob/master/dcos/http.py#L155


markusjura:
## Manual tests

**DC/OS cluster with auth enabled**

```
$ dcos config set core.dcos_url http://54.191.156.167
$ dcos auth login
...
$ dcos conduct info
ID  NAME  TAG  #REP  #STR  #RUN  ROLES
```

**DC/OS cluster with auth disabled**

```
$ cat etc/config-1.8.yaml
---
cluster_name: dcos-vagrant
bootstrap_url: http://boot.dcos
exhibitor_storage_backend: static
master_discovery: static
master_list:
- 192.168.65.90
resolvers:
- 10.0.2.3
superuser_username: admin
superuser_password_hash: "$6$rounds=656000$123o/Qz.InhbkbsO$kn5IkpWm5CplEorQo7jG/27LkyDgWrml36lLxDtckZkCxu22uihAJ4DOJVVnNbsz/Y5MCK3B1InquE6E7Jmh30" # admin
ssh_port: 22
ssh_user: vagrant
agent_list: []
check_time: false
oauth_enabled: 'false'
vagrant up

$ dcos config set core.dcos_url http://m1.dcos
$ dcos conduct info
UNLICENSED - please use "conduct load-license" to use more agents. Additional agents are freely available for registered users.
Licensed To: unknown
Max ConductR agents: 1
ConductR Version(s): 2.1.*
Grants: conductr, cinnamon, akka-sbr

ID  NAME  TAG  #REP  #STR  #RUN  ROLES
```

**ConductR standalone**

```
$ sandbox run 2.0.5
...
$ conduct info
ID               NAME              VER  #REP  #STR  #RUN  ROLES
bdfa43d-e5f3504  conductr-haproxy   v2     1     0     1  haproxy
7dc78f8          eslite             v1     1     0     1  elasticsearch
```

markusjura:
Note that once ConductR is part of the universe, we should we able to remove all DC/OS modules in the `conductr-cli`.

Issue #441, opened 2017-05-03
bndl endpoint ideas


longshorej:
I'm working on getting the maven build of chirper to work with ConductR's OCI support. This requires the use of `bndl --endpoint` and I'm glad to report that it works!

That being said, I have a couple ideas to make the `--endpoint` a little more user friendly.

1) Can we have it fail if the user specifies a `--component` flag for a component that doesn't exist (perhaps something like: `Invalid component hello3 (Available components: hello1, hello2)`
2) Can we make `--component` optional if the bundle only has a single component in it? For my usage, I have a bundle with a single component `oci-image`, I'd like to simply run `bndl --endpoint activity-stream --endpoint akka-remote` and, since there's only one component, have the command automatically know I'm referencing it.

WDYT @markusjura ?

markusjura:
> 1. Can we have it fail if the user specifies a --component flag for a component that doesn't exist (perhaps something like: Invalid component hello3 (Available components: hello1, hello2)

Yes, I'll create a PR for that today

> Can we make --component optional if the bundle only has a single component in it? For my usage, I have a bundle with a single component oci-image, I'd like to simply run bndl --endpoint activity-stream --endpoint akka-remote and, since there's only one component, have the command automatically know I'm referencing it.

Yes, I wanted to do that anyway. Will start with that today as well. The user doesn't need to specify the `--component` argument if either only one component in the `bundle.conf` exists or if two components exist and one of the components is the "status" component that contains the check start command.

@longshorej cc

PR #443, opened 2017-05-04
Auto-detection of endpoint component

markusjura:
Improves the `bndl --endpoint` argument by:
- Auto-detecting the endpoint component from the `bundle.conf` if the `--component` is not specified. Only works if one component in `bundle.conf` exist. If multiple components exist and the `--component` arg is not specified then a validation error is returned. The `bundle-status` component is excluded from the multiple endpoints list, i.e. is not counted.
- Adding a check if the `bundle.conf` contains at least one component. If not, a validation error is returned.
- Adding a check if the specified endpoint component exists in the `bundle.conf`. If not, a validation error is returned.

Fixes https://github.com/typesafehub/conductr-cli/issues/441.

### DIFF
--- a/conductr_cli/bndl_main.py
+++ b/conductr_cli/bndl_main.py
@@ -65,10 +65,6 @@ def set_endpoints(args):
         for endpoint_dict in args.endpoint_dicts:
             try:
                 endpoint = Endpoint(endpoint_dict)
-            except ValueError:
-                log.error('bndl: argument --component is required when specifying argument --endpoint {}'
-                          .format(endpoint_dict['name']))
-                sys.exit(2)
             except AmbigousBindProtocolError:
                 log.error('bndl: argument --bind-protocol is required '
                           'when acls with different protocol families are specified\n'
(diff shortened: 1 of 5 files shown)


markusjura:
Unit tests have been added for all the different conditions.

conductr_cli/bndl_main.py
```diff
@@ -127,18 +123,18 @@ def build_parser():
...
-                               metavar='ENDPOINT [OPTS]',
+                               metavar='ENDPOINT',
                                dest='endpoint_dicts',
                                default=[],
                                action=EndpointAction)
     endpoint_args.add_argument('--component',
                                help='Component to which an endpoint should be added\n'
                                     'Used in conjunction with the --endpoint option\n'
-                                    'Required when specifying the --endpoint option',
+                                    'When absent, the endpoints of the first component are replaced',
```
longshorej: The implementation is a bit different than the help text suggestions - it only autodetects if there is one component, it doesn't just use the first.

One idea would be to replace "When absent, the endpoints of the first component are replaced" with "Required when a bundle has more than one component"

conductr_cli/bndl_utils.py
```diff
@@ -192,15 +192,20 @@ def load_bundle_args_into_conf(config, args, with_defaults, validate_components)
...
+                endpoint_key = 'components.{}.endpoints'.format(component_name)
+                if endpoint_key in config:
+                    config.put(endpoint_key, None)
+            # Add endpoints to bundle components based on the --endpoint argument
+            for endpoint in args_endpoints:
+                component_name = detect_endpoint_component(config, endpoint)
+                endpoint_key = 'components.{}.endpoints'.format(component_name)
+                config.put(endpoint_key, endpoint.hocon())
+        elif validate_components:
+            raise SyntaxError('Unable to add endpoints. bundle.conf does not contain any component')
```
longshorej: "contain any components"

longshorej (approved):
Just a couple suggestions on wording but implementation looks great!

huntc:
Regarding adding a check component  if there's an endpoint... we don't want to always do that do we? The bundle may indeed have the smarts to have signalled its startup...

longshorej:
By check @markusjura meant validation, not the `check` command.

PR #444, opened 2017-05-04
WIP - DO NOT MERGE - OCI runtime bundle support

longshorej:
Getting what I have up on Github for now. This has gotten complicated because of ZIP files and symlinks. Need some more time to finish it up.


longshorej:
[Enhance the bndl tool to support OCI bundles](https://trello.com/c/f9Thaqev/603-enhance-the-bndl-tool-to-support-oci-bundles)

longshorej:
Maybe we'll come back to this someday but it's not a priority. ConductR supports the format so at least manually wiring this up is still possible using a hand-crafted `bundle.conf`. If demand is ever there we can look at adding it as a format supported by `bndl`.

PR #448, opened 2017-05-05
Add `--no-default-check` to `bndl`

longshorej:
This PR adds a flag, `--no-default-check` to `bndl`. The use-case would be a docker/oci image that uses `conductr-lib` and exposes ports. The user may want the default endpoints but not the default check command so with this flag we can support that case.

### DIFF
--- a/conductr_cli/bndl_main.py
+++ b/conductr_cli/bndl_main.py
@@ -121,6 +121,13 @@ def build_parser():
                         dest='use_default_endpoints',
                         action='store_false')
 
+    parser.add_argument('--no-default-check',
+                        help='If provided, a bundle will not contain a default check command\n'
+                             'For use with docker and oci-image formats',
+                        default=True,
+                        dest='use_default_check',
+                        action='store_false')
+
     endpoint_args = parser.add_argument_group('endpoints')
     endpoint_args.add_argument('-e', '--endpoint',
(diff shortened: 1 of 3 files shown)


huntc:
Given that our primary use case will be to support RP based projects with ConductR Lib, should we reverse things? Ie enable checks explicitly instead.

huntc:
If it is possible to auto-detect whether this is an RP app somehow then my question is answered. The presence of conductr-bundle-lib for a service could be a reasonable file to look for... although I'm wondering if this feels a bit magic.

longshorej:
If we reverse things, very few docker images will run out of the box. Because of this, I think many users will think ConductR is broken. I see where you're coming from though.

Inspecting files is a bit tricky - we'd have to peek inside the various layers for OCI images and iterate through every file in them. Plus magic as you said.

I'm not sure there's a clear winner here. Probably have to think about it/discuss it more.

markusjura:
The case that a docker-based service uses conductr-bundle-lib and has exposed a port in the `Dockerfile` explicitly is rather rare, correct? For example, a Dockerfile for a Lagom service, generated by fabric8 would not expose a port in the Dockerfile, correct?

If that is true then I think it is reasonable to use the flag as `--no-default-check` and to not apply additional magic.

PR #471, opened 2017-05-25
Explicitly encode/decode text files to UTF-8

huntc:
We need to be explicit about the encoding of text files otherwise the platform's locale will determine it, and potentially cause issues.


huntc:
Tested locally by removing license and auth tokens and going through that entire process. Additionally tested resolving of a Docker image. All appears well.

fsat (approved):

PR #498, opened 2017-06-19
Modify `conduct load-license` to treat 403 the same as 401 and 303 and thus delete auth-token file when invalid

### LINKED ISSUE #497: Prompt for the auth token when invalid
In the situation where the following occurs:

```
$ conduct load-license 
Error: Your credentials are no longer valid. Please visit https://www.lightbend.com/account/access-token to renew them.
```

... we should perhaps overwrite the credentials from `~/.lightbend/auth-token`. Not doing so prevents a new auth token from being applied. The above scenario should cause a prompt for the new auth token.

longshorej:
Fixes #497 

The thing I'm not clear on is why we were treating 403 differently to begin with. @fsat can you comment on that?

### DIFF
--- a/conductr_cli/license.py
+++ b/conductr_cli/license.py
@@ -44,11 +44,8 @@ def download_license(args, save_to, use_cached_auth_token):
     if log.is_verbose_enabled():
         log.verbose(response.text)
 
-    if response.status_code == 401 or response.status_code == 303:
+    if response.status_code == 401 or response.status_code == 303 or response.status_code == 403:
         license_auth.remove_cached_auth_token()
-        raise LicenseDownloadError([response.text])
-
-    elif response.status_code == 403:
         raise LicenseDownloadError([response.text])
 
     validation.raise_for_status_inc_3xx(response)


fsat:
@longshorej - I think `403` indicates user hasn't accepted T&C, hence we don't prompt for the token:

> An HTTP 403 (forbidden) status indicates that, while the user is able to be authenticated, they are unable to obtain a license given that they have not accepted Terms and Conditions. The following message should be output:
>
> Please accept our terms and conditions first. Please visit https://www.lightbend.com/account/access-token to accept them.

This is from the end of page 2 of the `Limited Use Production Suite Solution Design`:

https://docs.google.com/document/d/1zqHR6EcYJ1tnwS_3Xr-tT5wkM6netwu9iEV6xH2dvRM/edit#

huntc:
I don't know if the solution design caters for reality. Kiki had a bad access token on disk for some reason... we need to be able to cater for that. I had to have her explicitly remove the access token file.

fsat:
@huntc - we can use `conduct load-license -f` to force the prompt though?

```
192-168-1-5:pr-490 felixsatyaputra$ conduct load-license --help
usage: conduct load-license [-h] [--scheme SCHEME] [--host HOST] [-i IP]
                            [-p PORT] [--base-path BASE_PATH]
                            [--api-version {1,2}] [--offline]
                            [--settings-dir CLI_SETTINGS_DIR]
                            [--custom-settings-file CUSTOM_SETTINGS_FILE]
                            [--custom-plugins-dir CUSTOM_PLUGINS_DIR] [-q]
                            [-v] [-f]

optional arguments:
  -h, --help            show this help message and exit
...
  -f, --force           Always prompts for authentication token when specified,
                        use this option to change authentication token between different users
```

huntc:
OMG - I forgot about that... can we highlight that in our message then?

fsat:
@huntc - sure we can, I'll create a PR for this.

fsat:
Closed in lieu of #501 which highlights `conduct load-license -f` when license or token related error occurs.

Issue #509, opened 2017-06-27
Error when loading a bundle


ashutoshraina:
Repro Steps :
We have a running instance of conductr with loaded bundles.

$ conduct info

```
ID               NAME                            TAG  #REP  #STR  #RUN  ROLES
b532472-92feca5  conductr-haproxy              2.1.0     1     0     1  haproxy
cabaae7          visualizer                    2.1.0     1     0     0  web
57e432d          eslite                        2.1.0     1     0     1  elasticsearch
6273d7a          reactive-maps-backend-region            1     0     1  intranet
870ee7d          continuous-delivery           2.1.0     1     0     1  continuous-delivery
```

1) Download the bundle from the bundles endpoint and unzip it.
2) Try to load the bundle again

```
$ conduct load continuous-delivery-2.1.0/
Retrieving bundle..
Error: bndl: bundle.conf validation errors:
  The following properties are not allowed to be empty: components.continuous-delivery.endpoints.akka-remote.services
```

```
$ bndl continuous-delivery-2.1.0 -o output.txt
Error: bndl: bundle.conf validation errors:
  The following properties are not allowed to be empty: components.continuous-delivery.endpoints.akka-remote.services
```

The conf file 

`$ cat continuous-delivery-2.1.0/bundle.conf`

```
version              = "1"
name                 = "continuous-delivery"
compatibilityVersion = "3"
tags                 = ["2.1.0"]
annotations          = {}
system               = "continuous-delivery"
systemVersion        = "3"
nrOfCpus             = 0.1
memory               = 402653184
diskSpace            = 200000000
roles                = ["continuous-delivery"]
components = {
  continuous-delivery = {
    description      = "continuous-delivery"
    file-system-type = "universal"
    start-command    = ["continuous-delivery/bin/continuous-delivery", "-J-Xms134217728", "-J-Xmx134217728", "-Dhttp.address=$DEPLOYMENTS_BIND_IP", "-Dhttp.port=$DEPLOYMENTS_BIND_PORT", "-Dplay.crypto.secret=636f6e74696e756f75732d64656c6976657279"]
    endpoints = {
      "deployments" = {
        bind-protocol = "http"
        bind-port     = 0
        service-name  = "deployments"
        acls          = [
          {
            http = {
              requests = [
                {
                  path-beg = "/deployments/events"
                },
                {
                  path-beg = "/deployments"
                },
                {
                  path-beg = "/deployments/batches"
                },
                {
[...]

PR #511, opened 2017-06-28
Allow empty services and acls arrays

markusjura:
The bundle validation failed when an endpoint contained an empty array in the `services` or `acls` fields. However, empty fields are valid here, especially considering that a bundle configuration can declare those values to override existing `services` or `acls` fields with an empty array.

This change now allows empty `services` and `acls` fields.


markusjura:
Fixes https://github.com/typesafehub/conductr-cli/issues/509.

fsat (approved):

Issue #519, opened 2017-07-17
[Discussion] Consolidation of Protocol Requests


ashutoshraina:
We have api calls to conductr api in seeral different places. These can be consolidated together to form an sdk in a module to aid consistent use as well as re-use. This would make composing a lot of the functionality a lot easier. e.g. conduct load can be support -f flag #508 .

ashutoshraina:
addressed with #541

PR #527, opened 2017-07-21
extract control protocol http calls

ashutoshraina:
This PR extracts the load control protocol call. This is the first step in centralising the control protocol calls to enable re-use across different CLI functions.
This is also needed to facilitate re-use with upcoming restore functionality.
Relates to #519

### DIFF
--- a/conductr_cli/conduct_load.py
+++ b/conductr_cli/conduct_load.py
@@ -1,12 +1,11 @@
 from pyhocon import ConfigFactory, ConfigTree
 from pyhocon.exceptions import ConfigMissingException
-from conductr_cli import bndl_main, bundle_utils, conduct_request, conduct_url, constants, screen_utils, validation
+from conductr_cli import bndl_main, bundle_utils, constants, screen_utils, validation
 from conductr_cli.exceptions import MalformedBundleError, InsecureFilePermissions
 from conductr_cli import resolver, bundle_installation
 from conductr_cli.constants import DEFAULT_BUNDLE_RESOLVE_CACHE_DIR, \
     DEFAULT_CONFIGURATION_RESOLVE_CACHE_DIR
 from conductr_cli.bndl_utils import BndlFormat
-from conductr_cli.conduct_url import conductr_host
 from functools import partial
 from requests_toolbelt.multipart.encoder import MultipartEncoder, MultipartEncoderMonitor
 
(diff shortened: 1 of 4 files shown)


markusjura (approved):

conductr_cli/conduct_load.py
```diff
@@ -117,17 +117,9 @@ def load_v1(args):
...
-                                    verify=args.server_verification_file,
-                                    headers={'Content-Type': multipart.content_type})
-    validation.raise_for_status_inc_3xx(response)
-
-    if log.is_verbose_enabled():
-        log.verbose(validation.pretty_json(response.text))
+    response = load_bundle(args, multipart)
 
-    response_json = json.loads(response.text)
+    response_json = json.loads(response)
```
fsat: I think the `json.loads` should be done as part of `load_bundle` since we're displaying the JSON response in the `load_bundle` method if `log.is_verbose_enabled()`.

conductr_cli/conduct_load.py
```diff
@@ -291,22 +283,12 @@ def load_v2(args):
...
```
fsat: I think the `json.loads` should be done as part of `load_bundle` since we're displaying the JSON response in the `load_bundle` method if `log.is_verbose_enabled()`.

conductr_cli/conduct_stop.py
```diff
@@ -12,16 +12,9 @@ def stop(args):
...
-    response = conduct_request.put(args.dcos_mode, conductr_host(args), url, auth=args.conductr_auth,
-                                   verify=args.server_verification_file, timeout=DEFAULT_HTTP_TIMEOUT)
-    validation.raise_for_status_inc_3xx(response)
-
-    if log.is_verbose_enabled():
-        log.verbose(validation.pretty_json(response.text))
+    response = stop_bundle(args)
 
-    response_json = json.loads(response.text)
+    response_json = json.loads(response)
```
fsat: I think the `json.loads` should be done as part of `stop_bundle` since we're displaying the JSON response in the `stop_bundle` method if `log.is_verbose_enabled()`.

fsat (changes requested):
@ashutoshraina - nicely done! I'd suggest for `load_bundle` and `stop_bundle` to return the parsed json instead of plain text.

The reason for this suggestion is because these methods are already displaying the JSON when log verbose is enabled.

conductr_cli/conduct_stop.py
ashutoshraina: done

conductr_cli/conduct_load.py
ashutoshraina: done

conductr_cli/conduct_load.py
ashutoshraina: done

fsat (approved):

PR #529, opened 2017-07-25
extract members and agents calls.

ashutoshraina:
This PR extracts member and agent calls into the control protocol.
Reduces duplication as well.
Continues work from #527


fsat (approved):

longshorej (approved):
LGTM. I'd also like to consider the next step - should we be decoupling the control protocol methods from `args` altogether and simply pass in each argument that the method uses from the call sites, or is that too much boilerplate?

ashutoshraina:
I agree, there is no need to pass the entire args.
We should make that a different PR though.

PR #534, opened 2017-08-09
Remove base64 encoding of Bearer token for load-license OAuth requests

longshorej:
This removes the encoding of the `Bearer` token -- OAuth spec isn't as clear as it could be here, but it's intent is that this token is not base64 encoded, only that it adheres to the grammar laid out in `b64token`.

See this discussion on the spec that confirms this: https://www.ietf.org/mail-archive/web/oauth/current/msg08481.html , notably the following quotes:

> Note that the name b64token does not imply base64 encoding but rather
is the name for an ABNF syntax definition from HTTP/1.1, Part 7
[I-D.ietf-httpbis-p7-auth]. Because of its use, the "access_token"
string value from an access token response MUST match the b64token
ABNF so it can be used with the Bearer HTTP scheme.
-----------------------------------------
> Yeah, something as simple as, "Note that the name 'b64token' does not imply
> base64 encoding, see the definition in [[INSERT REFERENCE HERE]]." would do
> it.

### Manual Test

This was performed against the current website, which doesn't accept base64 encoded tokens.

```bash
$ conduct load-license

An access token is required. Please visit https://www.lightbend.com/platform/enterprise-suite/access-token to 
obtain one for free or for your commercial licenses.

Please enter your access token: <redacted>
Loading license into ConductR at 192.168.10.6

Licensed To: 521e4ec3-6e40-4719-8ec5-17e44262c41e
Max ConductR agents: 3
Expires In: 92 days (Thu 09 Nov 2017 09:37AM)
ConductR Version(s): 2.1.*, 2.1.0-beta.1
Grants: akka-sbr, cinnamon, conductr

License successfully loaded
-> 0
```

### DIFF
--- a/conductr_cli/license.py
+++ b/conductr_cli/license.py
@@ -3,7 +3,6 @@ from conductr_cli.conduct_url import conductr_host
 from conductr_cli.exceptions import LicenseDownloadError
 from dcos.errors import DCOSHTTPException
 import arrow
-import base64
 import datetime
 import json
 import logging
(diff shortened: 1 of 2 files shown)


huntc:
Nice pick up. However aren't we just passing back the access token that is input by the user? Surely, the change is at lightbend.com...

huntc:
Sorry, I now see. I didn't realise that we were encoding it here.

huntc (approved):

PR #544, opened 2017-08-25
Ensure correct field is used for DC/OS service name

longshorej:
Follow-up on #543 - I should be using `DCOS_SERVICE_NAME` instead of `id`.


huntc (approved):

huntc:
Please cherry pick

PR #546, opened 2017-08-31
Ensure that conduct_dcos.service_name is called

longshorej:
Looks like there was a regression when the restore PR was rebased -- the change to use `conduct_dcos.service_name` was reverted. This restores it.


ashutoshraina:
woops ..

ashutoshraina (approved):

